PR #26, opened 2021-07-28
Fixed small mypy issue when working with application_id

zomglings:
## Changes

Fixes mypy error where mypy couldn't understand that `MOONSTREAM_APPLICATION_ID` was not `Optional[str]` but rather a `str`.

## How to test these changes?

From `backend/` with virtualenv activated:
```
mypy moonstream/
```

Should exit 0


Yhtiyar (approved):
lgtm

PR #28, opened 2021-07-29
Created additional indices

kompotkot:
## Changes

Looks like Foreign key required index setup. Also added additional indices for feature.


zomglings:
Closing this because we need to change schemas (to make numeric values numeric).

PR #45, opened 2021-08-03
Improved speed for missing blocks command check

kompotkot:
## Changes

Now missing blocks check finishing in seconds.

### DIFF
--- a/crawlers/moonstreamcrawlers/cli.py
+++ b/crawlers/moonstreamcrawlers/cli.py
@@ -83,7 +83,7 @@ def ethcrawler_blocks_sync_handler(args: argparse.Namespace) -> None:
     starting_block: int = args.start
     while True:
         bottom_block_number, top_block_number = get_latest_blocks(
-            bool(args.transactions)
+            with_transactions=not args.notransactions
         )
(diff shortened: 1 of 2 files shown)


crawlers/moonstreamcrawlers/cli.py
```diff
@@ -207,8 +208,9 @@ def processing_order(raw_order: str) -> ProcessingOrder:
     parser_ethcrawler_blocks_sync.add_argument(
         "-t",
         "--transactions",
-        choices=["True", "False"],
-        default="True",
+        action="store_const",
```
zomglings, line 211: @kompotkot : Use [`action="store_true"`](https://docs.python.org/3/library/argparse.html#action).

If you want default to be `True`, use a flag like `--notransactions` or `--skip-transactions` with `action="store_true"`.

zomglings (approved):
lgtm

Left a small suggestion about using `"store_true"` instead of `"store_const"`. Change if you want to.

crawlers/moonstreamcrawlers/cli.py
kompotkot: Done

PR #49, opened 2021-08-03
Fixing txinfo

Yhtiyar:


backend/moonstream/data.py
```diff
@@ -94,8 +94,8 @@ class EthereumTransaction(BaseModel):
     gas: int
     gasPrice: int
     value: int
-    from_address: str = Field(alias="from")
-    to_address: Optional[str] = Field(default=None, alias="to")
+    from_address: str = Field(alias="from_address")
```
zomglings, line 97: @Yhtiyar : You don't need to put `alias="from_address"` if you are just using the same name as the member.

Why do you want to remove the alias? The alias allowed us to respect web3 schema for transactions.

backend/moonstream/data.py
```diff
@@ -94,8 +94,8 @@ class EthereumTransaction(BaseModel):
...
+    to_address: Optional[str] = Field(default=None, alias="to_adress")
```
zomglings, line 98: Typo in alias - `to_adress` -> `to_address`. Question about alias still remains.

zomglings (commented):
Just a small question about API request schema.

backend/moonstream/data.py
Yhtiyar: Because transaction object in frontend has `from_address` instead of `from`

zomglings (approved):
lgtm

PR #89, opened 2021-08-06
Improving txinfo

### LINKED ISSUE #82: Gas price and gas limit are confusing to users
We can present this information to users in a more intuitive way.

Ethereum developer docs, for example, have nice content explaining gas: https://ethereum.org/en/developers/docs/gas/

Yhtiyar:
## Changes
Changes `gas limit` label under `gas` in txinfo component to `amount of gas used in transaction`
Fixes bug when `currentTransaction` is null

## Related issues
 Resolves #82


zomglings (approved):
lgtm

PR #93, opened 2021-08-09
Ethscan crawler

Yhtiyar:
## Changes
Added etherscan.io crawler which gets sources of smart contracts via api, and puts into AWS S3 bucket


crawlers/mooncrawl/eth_scan.py
```diff
@@ -0,0 +1,129 @@
...
+            print(address)
+            querry_url = BASE_API_URL + f"&address={address}&apikey={ETH_SCAN_TOKEN}"
+            crawl_step(db_session, address, querry_url)
+            time.sleep(interval)
+        pass
+
+
+def load_smart_contracts() -> List[Tuple[str, str]]:
+    smart_contracts: List[Tuple[str, str]] = []
+    with open("moonstreamcrawlers/data/verified-contractaddress.csv") as csv_file:
```
zomglings, line 87: @Yhtiyar : You should not be storing this CSV in git.

crawlers/mooncrawl/eth_scan.py
```diff
@@ -0,0 +1,129 @@
...
+        csv_reader = csv.DictReader(csv_file)
+        for row in csv_reader:
+            smart_contracts.append((row["Txhash"], row["ContractAddress"]))
+    return smart_contracts
+
+
+crawl(load_smart_contracts(), 0.2)
```
zomglings, line 94: @Yhtiyar : This does not belong here. You should be doing this in `main`.

crawlers/sample.env
```diff
@@ -2,3 +2,7 @@
 export MOONSTREAM_IPC_PATH=null
 export MOONSTREAM_CRAWL_WORKERS=4
 export MOONSTREAM_DB_URI="postgresql://<username>:<password>@<db_host>:<db_port>/<db_name>"
+
+export ETH_SCAN_TOKEN="TOKEN"
```
zomglings, line 6: `MOONSTREAM_ETHERSCAN_TOKEN`

crawlers/sample.env
```diff
@@ -2,3 +2,7 @@
...
+export AWS_S3_SMARTCONTRACT_BUCKET=""
```
zomglings, line 7: AWS_S3_MOONSTREAM_SMARTCONTRACT_BUCKET

zomglings: I don't think you need a prefix. Prefix should be determined by crawler.

crawlers/mooncrawl/eth_scan.py
```diff
@@ -0,0 +1,129 @@
...
+    result_bytes = json.dumps(contract_data).encode("utf-8")
+    result_key = f"{contracts_prefix}/v1/{contract_file}"
+
+    s3 = boto3.client("s3")
+    s3.put_object(
+        Body=result_bytes,
+        Bucket=bucket,
+        Key=result_key,
+        ContentType="application/json",
+        Metadata={"smart_contract": "source"},
```
zomglings, line 40: Metadata should include `crawler_version` and `crawled_at`.

crawlers/mooncrawl/eth_scan.py
```diff
@@ -0,0 +1,129 @@
...
+            current_interval *= 2
+            time.sleep(current_interval)
+
+    if response is None:
+        print(f"Could not process URL: {crawl_url}", file=sys.stderr)
+        return None
+    page = response.json()
+    result = page["result"][0]
+    push_to_bucket(result, f"sources/{contract_address}.json")
+    # print(json.dumps(page["result"][0]))
```
zomglings, line 66: Remove unnecessary comment

crawlers/mooncrawl/eth_scan.py
```diff
@@ -0,0 +1,129 @@
...
+        except:
+            current_interval *= 2
+            time.sleep(current_interval)
+
+    if response is None:
+        print(f"Could not process URL: {crawl_url}", file=sys.stderr)
+        return None
+    page = response.json()
+    result = page["result"][0]
+    push_to_bucket(result, f"sources/{contract_address}.json")
```
zomglings, line 65: path should be: `/v1/etherscan/{contract_address}.json`.

crawlers/mooncrawl/eth_scan.py
```diff
@@ -0,0 +1,129 @@
...
+            success = True
+        except:
+            current_interval *= 2
+            time.sleep(current_interval)
+
+    if response is None:
+        print(f"Could not process URL: {crawl_url}", file=sys.stderr)
+        return None
+    page = response.json()
+    result = page["result"][0]
```
zomglings, line 64: Wrap the result in a JSON object like this:
```json
{
    "source": "etherscan",
    "crawler_version": "0.0.2",
    "crawled_at": "<time at which data was crawled>",
    "data": { ... }
}
```

zomglings (changes requested):

crawlers/mooncrawl/eth_scan.py
```diff
@@ -0,0 +1,184 @@
...
+    result_bytes = json.dumps(contract_data).encode("utf-8")
+    result_key = contract_file
+
+    s3 = boto3.client("s3")
+    s3.put_object(
+        Body=result_bytes,
+        Bucket=bucket,
+        Key=result_key,
+        ContentType="application/json",
+        Metadata={"source": "etherscan", "crawler_version": ""},
```
zomglings, line 50: Why is `crawler_version` empty?

crawlers/mooncrawl/eth_scan.py
```diff
@@ -0,0 +1,184 @@
...
+    object_key = f"/v1/etherscan/{contract.address}.json"
+    push_to_bucket(contract_info, object_key)
+
+    eth_address_id = get_address_id(db_session, contract.address)
+
+    eth_label = EthereumLabel(
+        label=ETHERSCAN_SMARTCONTRACTS_LABEL_NAME,
+        address_id=eth_address_id,
+        label_data={
+            "key": object_key,
```
zomglings, line 116: Instead of key, can you put the full `s3_uri` here?

`"s3_uri": f"s3://{bucket}/{object_key}`.

This way we are okay even if we change the bucket in the future.

zomglings: Later, we may even want to switch from AWS S3 to Google Cloud Storage. I don't mind if we use the `s3_uri` key for both, but if you want to be really future proof, call it something like `"object_uri"`.

crawlers/mooncrawl/eth_scan.py
```diff
@@ -0,0 +1,184 @@
...
+def crawl(
+    smart_contracts: List[VerifiedSmartContract],
+    interval: float,
+    start_step=0,
+):
+    with yield_db_session_ctx() as db_session:
+        for i in range(start_step, len(smart_contracts)):
+            contract = smart_contracts[i]
+            print(f"Crawling {i+1}/{len(smart_contracts)} : {contract.address}")
+            querry_url = (
```
zomglings, line 135: querry -> query

crawlers/mooncrawl/eth_scan.py
```diff
@@ -0,0 +1,184 @@
...
+from typing import Any, List, Optional, Tuple, Dict
+from dataclasses import dataclass
+from sqlalchemy.sql.expression import label, text
+from .version import MOONCRAWL_VERSION
+from moonstreamdb.models import EthereumAddress, EthereumLabel
+import requests
+
+
+ETH_SCAN_TOKEN = os.environ.get("MOONSTREAM_ETHERSCAN_TOKEN")
+if ETH_SCAN_TOKEN is None:
```
zomglings, line 21: Validation can live here. Other crawlers may not care if this token is set or not.

crawlers/mooncrawl/eth_scan.py
```diff
@@ -0,0 +1,184 @@
...
+from datetime import datetime
+from typing import Any, List, Optional, Tuple, Dict
+from dataclasses import dataclass
+from sqlalchemy.sql.expression import label, text
+from .version import MOONCRAWL_VERSION
+from moonstreamdb.models import EthereumAddress, EthereumLabel
+import requests
+
+
+ETH_SCAN_TOKEN = os.environ.get("MOONSTREAM_ETHERSCAN_TOKEN")
```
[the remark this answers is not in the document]
zomglings: Recommend changing the name to `MOONSTREAM_ETHERSCAN_TOKEN`. It's only one more character but is much more natural to `grep` for.

crawlers/mooncrawl/eth_scan.py
```diff
@@ -0,0 +1,184 @@
...
+
+def crawl(
+    smart_contracts: List[VerifiedSmartContract],
+    interval: float,
+    start_step=0,
+):
+    with yield_db_session_ctx() as db_session:
+        for i in range(start_step, len(smart_contracts)):
+            contract = smart_contracts[i]
+            print(f"Crawling {i+1}/{len(smart_contracts)} : {contract.address}")
```
zomglings, line 134: You could end up with `"Crawling 2001/1000: <address>"`, right?

Maybe better message here.

crawlers/mooncrawl/eth_scan.py
```diff
@@ -0,0 +1,184 @@
...
+            )
+            crawl_step(db_session, contract, querry_url)
+            time.sleep(interval)
+        pass
+
+
+def load_smart_contracts() -> List[VerifiedSmartContract]:
+    smart_contracts: List[VerifiedSmartContract] = []
+    s3 = boto3.client("s3")
+    data = s3.get_object(Bucket=bucket, Key="util/verified-contractaddress.csv")
```
zomglings, line 146: Define this key as a constant at the top of the file. Raise a specific error if this doesn't exist, as it indicates that there's something seriously wrong with our data infrastructure.

crawlers/mooncrawl/eth_scan.py
```diff
@@ -0,0 +1,184 @@
...
+        description="Crawls smart contract sources from ethersan.io"
+    )
+    parser.add_argument(
+        "--interval",
+        type=float,
+        default=0.2,
+        help="Number of seconds to wait between requests to the etherscan.io (default: 0.2)",
+    )
+    parser.add_argument(
+        "--offset",
```
zomglings, line 169: Good! You are thinking about checkpoints.

crawlers/mooncrawl/eth_scan.py
```diff
@@ -0,0 +1,184 @@
...
+                address=row["ContractAddress"],
+                name=row["ContractName"],
+            )
+        )
+    return smart_contracts
+
+
+def main():
+    parser = argparse.ArgumentParser(
+        description="Crawls smart contract sources from ethersan.io"
```
zomglings, line 160: ethersan -> etherscan

crawlers/mooncrawl/eth_scan.py
```diff
@@ -0,0 +1,184 @@
+import argparse
```
zomglings, line 1: Change filename to `etherscan.py`. Make it clear what the source is.

crawlers/setup.py
```diff
@@ -38,13 +38,15 @@
...
         "web3",
+        "boto3",
     ],
     extras_require={"dev": ["black", "mypy", "types-requests"]},
     entry_points={
         "console_scripts": [
             "ethcrawler=mooncrawl.ethcrawler:main",
             "esd=mooncrawl.esd:main",
             "identity=mooncrawl.identity:main",
+            "eth_scan=mooncrawl.eth_scan:main",
```
zomglings, line 49: `eth_scan` -> `etherscan`.

crawlers/mooncrawl/eth_scan.py
```diff
@@ -0,0 +1,184 @@
...
+        print(f"Could not process URL: {crawl_url}", file=sys.stderr)
+        return None
+    page = response.json()
+    result = page["result"][0]
+    contract_info = {
+        "data": result,
+        "crawl_version": MOONCRAWL_VERSION,
+        "crawled_at": f"{datetime.now()}",
+    }
+    object_key = f"/v1/etherscan/{contract.address}.json"
```
zomglings, line 107: `/etherscan/v1/{contract.address}.json` makes more sense to me. Then our bucket is organized by source. And versions will be different for different crawlers (etherscan vs. coinmarketcap).

crawlers/mooncrawl/eth_scan.py
zomglings: Might be good to address in the future:

If `crawl_step` fails to crawl a certain smart contract, then we are going to be looking through the logs for the print messages saying that you couldn't crawl a given contract and re-run the crawl only for those contracts.

The current setup where `--offset` is our only checkpoint recovery method will not work well there.

No need to fix in this PR, but please create an issue to improve checkpointing for this crawler.

zomglings (commented):
Left several comments + suggestions.

zomglings (approved):
lgtm

PR #98, opened 2021-08-12
Fix issues.

Andrei-Dolgolev:
1) issue with empty value in filter
2) issue with empty ethereum transaction query
Add
1) Add urlencoding and replace null to 0 for timestamps
2) Add sorting by timestamps


zomglings (approved):
lgtm

PR #99, opened 2021-08-13
Cards improvement

### LINKED ISSUE #61: First round of feedback on Moonstream frontend
## Transaction hashes

Would be nice to see transaction hashes on the cards in the stream. These are quite commonly used to verify that you're looking at the right transaction (across sites, from Javascript console, etc.). It should be prominent.

(Currently, only shows up in transaction details.)

## Colors

Cards feel a little too colorful. Would prefer that colors are used to display information about subscriptions only. Rest of the card can be white or dark depending on theme.

Examples to consider:
1. @kompotkot : Suggested this site - https://dribbble.com/search/card%20component%20list
2. @Andrei-Dolgolev : liked this for minimalism - https://dribbble.com/shots/15221832/attachments/6968098?mode=media
3. @zomglings : Likes the GitHub repository cards that show up in Slack - 
image

## Transaction details should not take over full view

Would be nice to be able to switch between transactions when looking at details. This could be similar to entries list vs. entries on [Bugout frontend](https://bugout.dev).

@kompotkot can fill in more details if this ever becomes confusing.

### LINKED ISSUE #81: Green arrow to expand transaction details is easy to miss
It could be because the transaction cards are too colorful, but users have missed the green arrow element on the frontend completely during onboarding. Maybe we should make it more obvious.

peersky:
## Changes

- Cards colors
- Added nonce info in card
- Cards bottom is now flexbox
- Changed detail button color to `secondary`
- In Desktop view list of cards takes now 55% of screen, rest 45% takes side view 
- Sideview when no entry is selected is onboarding experience
- Added colors matching with active users subscription address colors. It will appear as background of an addresss 
- Simplified UI context to monitor state of UI in `/steam`: switching any state happens now via `setEntryDetailView`, when loading new transaction just`setCurrentTransaction`.  `setEntriesViewMode` is not exported anymore and being managed by internal UI context useEffects. 
-  Removed `entryId` from UI context as it's not being used. If need to get entryId from context - use `currentTransaction` object to set/get it
- Disabled split-pane resizer in `/stream` view 
- Added hash in to Ethereum cards
- Entry details are not any more an URL page. Instead moved it in to `/components/SteamEntryDetails.js`. it's now called from `/stream` page. 

## How to test these changes?

Tested locally on `chrome` `safari` `firefox`

## Related issues

resolves #61
resolves #81

### DIFF
--- a/frontend/pages/stream/index.js
+++ b/frontend/pages/stream/index.js
@@ -1,6 +1,49 @@
+import React, { useContext } from "react";
 import { getLayout } from "../../src/layouts/EntriesLayout";
+import StreamEntryDetails from "../../src/components/SteamEntryDetails";
+import UIContext from "../../src/core/providers/UIProvider/context";
+import {
+  Box,
+  Heading,
+  Text,
+  Stack,
+  UnorderedList,
+  ListItem,
+} from "@chakra-ui/react";
 const Entry = () => {
-  return "";
+  const ui = useContext(UIContext);
+
+  // return "";
+  if (ui.currentTransaction) {
+    return <StreamEntryDetails />;
+  } else
+    return (
+      <Box px="7%" pt={12}>
+        <>
+          <Stack direction="column">
+            <Heading>Stream view</Heading>
+            <Text>
+              In this view you can follow events that happen on your subscribed
+              addresses
+            </Text>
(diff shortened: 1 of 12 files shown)


zomglings (approved):
lgtm

PR #104, opened 2021-08-13
Fixed issues with stream requests to backend

zomglings:
## Changes

Axios was encoding query parameters and we were doing so as well. Fixed it so that we rely on Axios behaviour.


Andrei-Dolgolev (approved):
Looks correct

PR #105, opened 2021-08-15
Allow subscriptions to Ethereum whale watch events

### LINKED ISSUE #76: "/subscription/types" should return "subscription_type"
Currently, there is some confusion because the backend removed the `"subscription_plan_id"` field, which the frontend was using to determine which icon to put next to each subscription on the `/subscriptions` view.

We should just use a `"subscription_type"` field, with possible values: `ethereum_blockchain`, `ethereum_txpool`, etc.

### LINKED ISSUE #83: Show addresses that are experiencing high transaction volume as suggested subscriptions
One of the biggest problems with the subscription model is that, if the user does not know addresses to subscribe to, they have an empty stream.

A really good idea that one of our early users had was to show addresses that were experiencing very high transaction volume within the last `x` hours.

This could be its own subscription - `ethereum_trending`. Our subscription model might have to change slightly, as this subscription would not take an address, but it would have a color and a label.

This is a fantastic idea and is a good way to implement an always on default stream when a user signs up. High priority to add this.

zomglings:
## Changes

This PR introduces the notion of multiple event providers for a single stream. It is the basis for generalizing from just Ethereum transactions to sources like the Ethereum transaction pool, other blockchains, etc.

As such, it is very involved and spans the frontend, the backend, and the crawlers.

## How to test these changes?

Manually for now.

## Related issues

Resolves #83 
Resolves #76


zomglings:
Closing in favor of https://github.com/bugout-dev/moonstream/pull/137

PR #110, opened 2021-08-17
Fix stream loading

Andrei-Dolgolev:
## Changes
Switch stream sql request to create one union table.
Add correct pagination for useQuery.
Add some fixes on pagination logic and disable load more.


zomglings:
Closed in favor of #137

PR #117, opened 2021-08-18
 entry details will better fit in to split view

peersky:
## Changes

Made table smaller, hash title smaller, added `word-break: all` to table rows so long rows like addresses wrap on smaller screens

## How to test these changes?

localhost `chrome`


zomglings (approved):
lgtm

PR #135, opened 2021-08-20
split stream card to ethereum card and wrapper

### LINKED ISSUE #134: Support different card components in stream
We need common wrapper for all cards components that we can reuse

peersky:
## Changes

/components: split `StreamEntry.js` in to `StreamEntry.js` and `stream-cards/EthereumMempool.js.

We can use same wrapper of `StreamEntry.js` and define new cards in `components/stream-cards`

## How to test these changes?

localhost `chrome`

## Related issues

resolves #134


zomglings (approved):
lgtm

PR #137, opened 2021-08-22
Introduce the concept of event providers for a stream, add an Ethereum whale watch event provider

### LINKED ISSUE #76: "/subscription/types" should return "subscription_type"

### LINKED ISSUE #83: Show addresses that are experiencing high transaction volume as suggested subscriptions

zomglings:
## Changes

This PR introduces the notion of multiple event providers for a single stream. It is the basis for generalizing from just Ethereum transactions to sources like the Ethereum transaction pool, other blockchains, etc.

As such, it is very involved and spans the frontend, the backend, and the crawlers.

## How to test these changes?

Manually for now.

## Related issues

Resolves #83 
Resolves #76


zomglings:
Opened this as a pull request from `bugout-dev/moonstream` because I updated the `moonstreamdb` version and wanted the `moonstreamdb` commit to be available on the main repo.

zomglings:
TODO(@zomglings ):
- [ ] Make Ethereum whale watch cards display the meaning of the statistics instead of `To:`. Maybe put a section title over each row.
- [ ] Amounts are currently displayed in WEI. Use @Yhtiyar 's utility method to convert to ETH.

frontend/src/core/hooks/useStream.js
```diff
@@ -1,61 +1,284 @@
...
+      onSuccess: (newEvents) => {
+        if (newEvents && newEvents.stream_boundary && newEvents.events) {
+          setEvents([...newEvents.events]);
+          updateStreamBoundaryWith(newEvents.stream_boundary);
+        }
       },
     }
   );
 
+  const { refetch: loadOlderEvents, isFetching: loadOlderEventsIsFetching } =
```
peersky, line 109: @zomglings I think you should not duplicate query in line 138

peersky: Instead I think would be cleaner to define 

{refetch} = useQuery(....)

const getOlder() = () => refetch({ direction: `older` }) 
const getOlder() = () => refetch({ direction: `newer` })

zomglings: 👍 will work on this today.

zomglings:
TODO(@zomglings ):
- [ ] Production migration plan for existing subscriptions and subscription_types.

backend/moonstream/routes/streams.py
```diff
@@ -56,54 +62,205 @@
...
+    if q.strip() != "":
+        query = stream_queries.parse_query_string(q)
+
+    events = latest_events(
+        db_session,
+        bc,
+        MOONSTREAM_DATA_JOURNAL_ID,
+        MOONSTREAM_ADMIN_ACCESS_TOKEN,
+        query,
+        1,
```
Andrei-Dolgolev, line 164: Still not sure must we limit ability of `num_events` parameter. right now it return one last element.

Andrei-Dolgolev (approved):
Looks correct

frontend/src/core/hooks/useStream.js
zomglings: Experimented with this and decided to keep things as they are. I might be missing something about `react-query` and if so I am happy to make the change in a future PR.

What I discovered. According to the [`useQuery` API](https://react-query.tanstack.com/reference/useQuery):
1. `refetch` always has the signature: `(options: {throwOnError: boolean, cancelRefetch: boolean}) => Promise<UseQueryResult>`. This means that it is not possible to pass arguments to `getEvents` using `refetch`.
2. The `queryFn` passed to `useQuery` (which ultimately calls `getEvents`) must have the signature: `(context: QueryFunctionContext) => Promise<TData>`. `QueryFunctionContext` does not give an easy way to pass arguments to the `getEvents` call that the `queryFn` must make.

This means that the only option left is to define `getEvents` as:
```js
const getEvents = (mode) => {
    // do some mode validation
    return async () => {
        // mode-specific logic to create stream boundary for request
        // perform request and return response
    };
}
```

And then we would have to make 3 separate `useQuery` calls anyway (using `getEvents("default")` or `getEvents("newer")` or `getEvents("older")` as the `queryFn` for each call, each with pretty different `onSuccess` handlers.

Removing the query-specific logic from where we define the queries and moving it to `getEvents` makes things a lot more confusing when reading the code.

@peersky: Please do let me know if there's a better way. But let's save it for a separate PR.

Andrei-Dolgolev:
Some proposals
https://github.com/bugout-dev/moonstream/issues/153

frontend/src/core/hooks/useStream.js
peersky: @zomglings I think refetch here won't work, my mistake. What would work is to define fetch function with an argument. 

Here is example how it's done in bugout search: 


```js
const useJournalEntries = ({
  journalId,
  journalType,
  isContent,
  pageSize,
  searchQuery,
}) => {
  const limit = pageSize ? pageSize : 25;

  const { entriesSearch } = useEntriesSearch({
    journalId,
  });

  const getEntries = (searchTerm) => async (key, jorunalId, nextPage = 0) => {
    if (!nextPage) {
      nextPage = 0;
    }

    const searchTags = searchTerm.split(" ").filter(function (n) {
      if (n.startsWith("#")) return n;
      else {
        return null;
      }
    });

    const data = await entriesSearch({
      searchTerm,
      journalType,
      isContent,
      limit,
      offset: nextPage,
      searchTags,
    });

    const newEntryList = data.data.results.map((entry) => ({
      ...entry,
      id: entry.entry_url.split("/").pop(),
    }));
    return {
      data: [...newEntryList],
      nextPage: nextPage + 1,
      next_offset: data.data.next_offset,
      total_results: data.data.total_results,
      offset: data.data.offset,
    };
  };

  const {
    data: EntriesPages,
    isFetchingMore,
    isLoading,
    canFetchMore,
    fetchMore,
    refetch,
  } = useInfiniteQuery(
    ["journal-entries", { journalId }],
    getEntries(searchQuery),
    {
      ...queryCacheProps,
      getNextPageParam: (lastPage) => lastPage.next_offset ?? false,
      getFetchMore: (lastGroup) => {
        return lastGroup.next_offset === null ? false : lastGroup.next_offset;
      },
      enabled: !!journalId,
    }
  );
```

Here `searchTerm` is being passed as argument.

Implementation for moonsteam would look like 

```js
// /component code 
const [fetchMethod, setFetchMethod] = useState('newest')
const streamCache  = useStream(otherArguments, {fetchMethod})

// hook code 
const useJournalEntries = ({
 fetchMethod,
  journalId,
  journalType,
  isContent,
  pageSize,
  searchQuery,
}) => {

...
useQuery(
    ["stream-entries", queryKeys],
    getStream(fetchMethod),
    .... )

```

In this implementation actually I don't see any problem to store fetchMethod inside queryKeys with name `lastFetchMethod: <Method>` , by that way avoiding need in both function wrapping and extra query hooks.

zomglings:
> TODO(@zomglings ):
> 
> * [x]  Make Ethereum whale watch cards display the meaning of the statistics instead of `To:`. Maybe put a section title over each row.
> * [x]  Amounts are currently displayed in WEI. Use @Yhtiyar 's utility method to convert to ETH.

Decided to keep amounts in WEI for now, but write that they are in WEI.

backend/moonstream/providers/bugout.py
```diff
@@ -0,0 +1,273 @@
...
+        timeout: float = 30.0,
+    ):
+        """
+        Args:
+        - event_type: Name of event this instance provides
+        - tags: Tags which define events that this provider works with
+        - batch_size: Number of events to read from journal at a time
+        - timeout: Request timeout for Bugout requests
+        """
+        self.event_type = event_type
```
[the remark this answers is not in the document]
zomglings: @Andrei-Dolgolev : I think this is a good idea - we can add the `token` and `journal_id` as arguments to `BugoutEventProvider`. This will make it easy for third parties to offer data through the Moonstream dashboard.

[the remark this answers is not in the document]
zomglings: Created an issue for this: https://github.com/bugout-dev/moonstream/issues/157

backend/moonstream/routes/streams.py
zomglings: It returns one element per provider, and handler decides how to deal with that.

frontend/src/core/hooks/useStream.js
zomglings: @peersky : I think that's the same as the final option from my comment above. I tried it and in the `useStream` case it really doesn't work well because the different `onSuccess` implementations still make it difficult to work with a single query.

See the order in which each `onSuccess` adds events to the `events` state.

peersky (approved):
lgtm

frontend/src/core/hooks/useStream.js
peersky: Using `switch` in `onSuccess` would not be reliable as it's async. I agree that in this case its lgtm to use multiple queries. 
However it still seems quite confusing. I made an issue to follow this topic: 
https://github.com/bugout-dev/moonstream/issues/158

PR #167, opened 2021-08-24
onboarding stream sceen StreamEntry card props

peersky:
## Changes

Due to changes in stream entry schema component was broken

## How to test these changes?

localhost `chrome`


zomglings (approved):
lgtm

PR #174, opened 2021-08-25
Whale watch card improvement

### LINKED ISSUE #173: Whale watch improvement
We want to make Whale Watch feature more easy to use. For that we want to implement following: 

- [x] New Subscription modal should not ask for `address` and `name` for whale watch 
- [ ] Backend should not allow to create more then one whale watch per token
- [x] #175
- [ ] We should provision that whale watch cards can be subscribed to different sources... Speaking in terms of whales, we can call this - `oceans`. That is platforms or protocols - ethereum, or twitter etc

### LINKED ISSUE #175: We should be able to open new subscription window to subscribe to addresses that are shown directly in stream whale watch card

peersky:
## Changes

- reworked whale watch card in to simple grid components
- added subscribe to address button if whale watch address is not subscribed yet
- Added initial properties to newSubscription modal that allow to predefine address and type
- Fixes bug with not being able to create new subscription from welcome screen
- Disabled entry details button for whale watch card as it has no details feature right now
- Changed time format to `<Date> from <time> to <time>` if new time date is different format will show as  `<Date> from <time> to <Date> <time>` 

## How to test these changes?

localhost `chrome`

## Related issues

resolves #175

### DIFF
--- a/frontend/src/components/ContractSource.js
+++ b/frontend/src/components/ContractSource.js
@@ -1,29 +1,29 @@
 import React from "react";
-import {
-    VStack,
-    Code,
-} from "@chakra-ui/react";
+import { VStack, Code } from "@chakra-ui/react";
 import { Table, Thead, Tbody, Tr, Th, Td } from "@chakra-ui/react";
 
 const ContractSource = ({ source_info }) => {
-    return (
-        <VStack spacing={3}>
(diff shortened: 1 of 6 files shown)


zomglings (approved):
lgtm

PR #176, opened 2021-08-25
Fixed transaction information service on frontend

### LINKED ISSUE #170: TxInfo details should respect new response body on "/streams" endpoint
Currently, frontend is not showing any transaction information.

This bug was caused by @zomglings 's PR in which he changed response structures so that event transaction information shows up under the ".event_data" field.

zomglings:
Also fixed request body on `/txinfo/ethereum_blockchain`.

## Changes

This PR fixes the issue introduced by the event providers PR in which transaction information was not displaying on the stream entry details view.

It also fixes an inconsistency in the `/txinfo/ethereum_blockchain` API endpoint about the casing of the `gas_price` paramter in the request body.

Finally, it adds some null checks on the `ui` state. This is because the edited lines were causing errors in the frontend during development and could potentially cause errors in production as well.

## How to test these changes?

Run locally.

## Related issues

Resolves #170

### DIFF
--- a/backend/moonstream/data.py
+++ b/backend/moonstream/data.py
@@ -88,10 +88,10 @@ class ContractABI(BaseModel):
 
 class EthereumTransaction(BaseModel):
     gas: int
-    gasPrice: int
+    gas_price: int
     value: int
-    from_address: str
-    to_address: Optional[str]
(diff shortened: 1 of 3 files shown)


Yhtiyar (approved):
lgtm

PR #177, opened 2021-08-25
Allow any subscription type to be created from welcome screen 

peersky:
## Changes

Subscriptions created from onboaring view are treated not as free subscriptions anymore 

## How to test these changes?

localhost `chrome`


zomglings (approved):
lgtm

PR #201, opened 2021-08-30
fixed address_info response schema

### LINKED ISSUE #194: /address_info/ethereum_blockchain should use key-value pairs to add information from different sources
Currently, we have an `"address_type"` field which takes values from the `AddressType` enum and `"details"` which contains different information depending on `"address_type"`.

A better schema would be:
```json
{
    "address": "<address>",
    "<address_type_1>": {...},
    ...,
    "<address_type_n>": {...}
}
```

This would allow us to use schema custom to each address type (e.g. `"token", "nft", etc.) and make it easier to parse this information on the frontend and in other clients.

Yhtiyar:
## Changes
Changed `address_info/` response schema

## How to test these changes?
manually

## Related issues
Resolves #194

### DIFF
--- a/backend/moonstream/actions.py
+++ b/backend/moonstream/actions.py
@@ -51,15 +51,6 @@ def get_contract_source_info(
     return None
 
 
-class AddressType(Enum):
-    UNKNOWN = "unknown"
-    REGULAR = "regular"
-    TOKEN = "token"
-    SMART_CONTRACT = "smart_contract"
-    NFT = "nft"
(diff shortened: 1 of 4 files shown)


Yhtiyar:
Example requests:
`localhost:7481/address_info/ethereum_blockchain?address=0x2e98a6804e4b6c832ed0ca876a943abd3400b224`

Expected responcse:

zomglings (approved):
lgtm

PR #220, opened 2021-09-01
Humbug reporter for moonstream backend

kompotkot:
## Changes

Gumbug reporter for backend.

## How to test these changes?

Tested manually.


kompotkot:
@bugout-dev check
- [ ] Add env var `HUMBUG_REPORTER_BACKEND_TOKEN`

backend/moonstream/routes/streams.py
```diff
@@ -121,17 +123,27 @@ async def stream_handler(
...
+            MOONSTREAM_ADMIN_ACCESS_TOKEN,
+            stream_boundary,
+            query,
+            user_subscriptions,
+            result_timeout=10.0,
+            raise_on_error=True,
+        )
+    except ReceivingEventsException as e:
+        logger.error("Error receiving events from provider")
+        reporter.error_report(e)
```
zomglings, line 140: From the API, can we report in `background_task`? It doesn't make sense to require the user to wait for our error reporting to complete before they get a response.

zomglings (changes requested):
All reporting done in API endpoints should report errors in a background task.

One thing you can consider is adding a middleware to all API endpoints. If handler raises exception, report in background and reraise.

backend/moonstream/routes/streams.py
zomglings: Actually, I think if you instantiate the reporter with default configuration, it sends reports in a thread.

Use [`mode=DEFAULT`](https://github.com/bugout-dev/humbug/blob/main/python/humbug/report.py#L60) and make sure `wait=False` on report calls. I think this is okay in your current code.

zomglings: Background tasks not needed, but I do recommend middleware.

kompotkot: Extended class `HTTPException` and `Response`

zomglings (approved):
lgtm

PR #237, opened 2021-09-06
Nft dashboard

peersky:


frontend/pages/nft.js
```diff
@@ -0,0 +1,86 @@
...
+          h="auto"
+          direction="row"
+          minW="240px"
+          spacing={[2, 0, null]}
+          boxShadow="md"
+          borderRadius="lg"
+          bgColor="gray.100"
+        >
+          <StatsCard
+            labelKey="transactions"
```
zomglings, line 62: This looks great.

frontend/src/components/StatsCard.js
```diff
@@ -0,0 +1,207 @@
...
+    { value: 1e6, symbol: "M" },
+    { value: 1e9, symbol: "G" },
+    { value: 1e12, symbol: "T" },
+    { value: 1e15, symbol: "P" },
+    { value: 1e18, symbol: "E" },
+  ];
+  const rx = /\.0+$|(\.[0-9]*[1-9])0+$/;
+  var item = lookup
+    .slice()
+    .reverse()
```
zomglings, line 38: Why not just create them in the correct order? (and put a comment indicating that order matters).

frontend/src/components/StatsCard.js
```diff
@@ -0,0 +1,207 @@
...
+import { TriangleDownIcon, TriangleUpIcon } from "@chakra-ui/icons";
+import useNFTs from "../core/hooks/useNFTs";
+import web3 from "web3";
+
+const TIME_PERIOD = {
+  current: 0,
+  previous: 1,
+};
+
+const isNumberNotZero = (str) => {
```
zomglings, line 12: Not a good name for this function. Maybe `isNumberNonzeroAndFinite`?

zomglings: Just do `return !(isNaN(Number(str)) || Number(str) === 0);`

frontend/src/components/StatsCard.js
```diff
@@ -0,0 +1,207 @@
...
+    { value: 1, symbol: "" },
+    { value: 1e3, symbol: "k" },
+    { value: 1e6, symbol: "M" },
+    { value: 1e9, symbol: "G" },
+    { value: 1e12, symbol: "T" },
+    { value: 1e15, symbol: "P" },
+    { value: 1e18, symbol: "E" },
+  ];
+  const rx = /\.0+$|(\.[0-9]*[1-9])0+$/;
+  var item = lookup
```
zomglings, line 36: It's not a good idea to use `item` as the name of the variable as well as the name of the argument to the callback in `find`.

zomglings: Also, as a general principle in this codebase, use `let` instead of `var`. `var` scoping can be confusing. `let` is clear.

Reference: https://stackoverflow.com/a/11444416

frontend/src/components/StatsCard.js
```diff
@@ -0,0 +1,207 @@
...
+    .slice()
+    .reverse()
+    .find(function (item) {
+      return num >= item.value;
+    });
+  return item
+    ? (num / item.value).toFixed(digits).replace(rx, "$1") + item.symbol
```
zomglings, line 43: Can you do this using numerical operations rather than string operations?

zomglings: I really don't like use of regex for this kind of thing. If you must use it, put a comment explaining what the regex is doing with some examples.

zomglings: Most likely, you do not have to use regex.

frontend/src/components/StatsCard.js
```diff
@@ -0,0 +1,207 @@
...
+    : "0";
+};
+
+const getChange = (a, b) => {
+  if (isNumberNotZero(a) && isNumberNotZero(b)) {
+    let retval = (Math.abs(Number(a) - Number(b)) * 100) / Number(b);
+    retval = retval > 9999 ? nFormatter(retval, 3) : retval;
```
zomglings, line 50: What if `retval` is negative? Maybe you should use `Math.abs()`?

frontend/src/components/StatsCard.js
```diff
@@ -0,0 +1,207 @@
...
+    return "-";
+  } else {
+    return Number(a) - Number(b);
+  }
+};
+
+const getSign = (a) => {
+  if (isNaN(a)) return "-";
+  return Number(a) >= 0 ? true : false;
+};
```
zomglings, line 68: Whitespace between function and component. (Prettier should fix this.)

frontend/src/components/StatsCard.js
```diff
@@ -0,0 +1,207 @@
...
+        <Stack
+          w="100%"
+          direction="row"
+          fontSize="1.125rem"
+          placeContent="center"
+          alignItems="center"
+        >
+          {nftData.isValueIncrease && <TriangleUpIcon color="suggested.900" />}
+          {!nftData.isValueIncrease && <TriangleDownIcon color="unsafe.900" />}
+          <Text
```
zomglings, line 154: Add `id` attributes to each of these tags so that we can understand easily in the code and in the DOM what the nodes represent.

Maybe some comments could help as well, to map the components to the picture of the card.

zomglings (commented):
Initial review after refactoring of the `Analytics` component.

frontend/src/components/StatsCard.js
peersky: because then it won't be `self-documenting`?

frontend/src/components/StatsCard.js
peersky: This regex part is to remove trailing zero. 

I decided to remove it at all as it keeps things clear on decimal parts. And also use 2 fractional digits instead of 3, seems visually more clear then

frontend/src/components/StatsCard.js
peersky: Good point

peersky: Also removed .toFixed from return as this can break in case when retval > 9999 is true and nFormatter returns a string. 

```js
retval =
      Math.abs(retval) > 9999 ? nFormatter(retval, 2) : retval.toFixed(2);
    return retval;
```

frontend/src/components/StatsCard.js
peersky: also renaming in to `isZeroOrPositive`

PR #247, opened 2021-09-09
HOME path missing when building with go deployment script

kompotkot:
## Changes

When instance execute startup script it missing $HOME variable and because of it also missing $GOPATH.

## How to test these changes?

Tested


Andrei-Dolgolev (approved):
Looks correct

PR #265, opened 2021-09-16
Added /streams/info endpoint

zomglings:
## Changes

This gives information about available event providers on the API. It is a public endpoint and does not require authentication.

## How to test these changes?

Run the API and:

```bash
curl localhost:7481/streams/info | jq .
```


Andrei-Dolgolev (approved):
Looks correct

PR #269, opened 2021-09-21
Option to run block sync from current block at Ethereum

kompotkot:
## Changes

Syncronization command to add blocks with 6 blocks delay is:
```python
python -m mooncrawl.ethcrawler blocks synchronize -c 6 -j 1
```

## How to test these changes?

Tested manually


Yhtiyar (approved):
lgtm

PR #276, opened 2021-09-23
Token screen

peersky:
## Changes

- Improved table layout for mobile view
- renamed notes to labels 
- made color separation of placeholder of unnamed token id labels to labels
- fixed bug on backend API that caused to add label to token that was passed in authorisation header, not one in the form
- fixed logic to hide frontend token from the table

## How to test these changes?

localhost, local dev server


kompotkot (approved):
lg

PR #282, opened 2021-09-24
Threads instead of pools, missing block cmd check num of transactions

kompotkot:
## Changes

While we have a troubles with Pool execution calls with psycopg2, using Threads executor instead.
Links:
- https://docs.sqlalchemy.org/en/14/core/pooling.html#sqlalchemy.pool.AssertionPool
- https://stackoverflow.com/questions/39493785/sqlalchemy-threadpoolexecutor-too-many-clients
- https://stackoverflow.com/questions/68609796/sqlalchemy-exc-operationalerror-psycopg2-operationalerror-ssl-error-wrong-ve
- https://medium.com/@philamersune/fixing-ssl-error-decryption-failed-or-bad-record-mac-d668e71a5409

## How to test these changes?

localy

### DIFF
--- a/crawlers/mooncrawl/mooncrawl/ethcrawler.py
+++ b/crawlers/mooncrawl/mooncrawl/ethcrawler.py
@@ -15,7 +15,6 @@ import dateutil.parser
 
 from .ethereum import (
     crawl_blocks_executor,
-    crawl_blocks,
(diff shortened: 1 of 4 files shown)


Andrei-Dolgolev (approved):
looks correct

PR #283, opened 2021-09-24
added reporter to nft crawler, removed global cache for addresses

Yhtiyar:
## Changes
Fixed nft crawler, it was not saving labels if there was problem with one contract name/symbol (unicode escape etc.)
added reports with batch information to fix problems later


zomglings (approved):
lgtm

We are not caching address IDs for now to protect ourselves against errors that we do not yet know about.

PR #304, opened 2021-10-05
Add ability copy sqlite3 database with filter and add dirive.

Andrei-Dolgolev:


datasets/nfts/nfts/cli.py
```diff
@@ -38,6 +40,49 @@ def handle_import_data(args: argparse.Namespace) -> None:
         import_data(target_conn, source_conn, event_type, args.batch_size)
 
 
+def handel_generate_filtered(args: argparse.Namespace) -> None:
```
zomglings, line 43: `handle_filter_data`.

datasets/nfts/nfts/cli.py
```diff
@@ -38,6 +40,49 @@ def handle_import_data(args: argparse.Namespace) -> None:
...
+
+    with contextlib.closing(sqlite3.connect(args.source)) as source_conn:
+
+        if not args.target:
```
zomglings, line 47: Get rid of this. Let us make `--target` required.

datasets/nfts/nfts/datastore.py
```diff
@@ -433,3 +433,37 @@ def import_data(
     if source_offset is not None:
         delete_checkpoints(target_conn, event_type, commit=False)
         insert_checkpoint(target_conn, event_type, source_offset)
+
+
+def filter_data(sqlite_db: sqlite3.Connection, cli_args):
```
zomglings, line 438: Make this signature more clear. I would expect a signature like this:

```python
def filter_data(conn: sqlite3.Connection, start_time: Optional[int] = None, end_time: Optional[int] = None) -> None:
    ...
```

This makes it so much more clear what this function is doing. In general, don't use these arbitrary argument objects like `cli_args`. The whole reason we have a handler function in the cli module is to convert from `args` object to meaningful arguments.

zomglings (approved):
lgtm

PR #350, opened 2021-10-31
Improve loading speed

peersky:
## Changes

After demoing to people Ive noticed some slow loading, after short investigation appears that using dynamic imports was bad idea for `<head>` tag generation within Next SSG compilation step. I removed dynamic imports and as well some of unused preload components 
Also, not git related, but still worth to mention - Reduced size of product page images on AWS . 

## How to test these changes?

Check in production your loading time

### DIFF
--- a/frontend/pages/_app.js
+++ b/frontend/pages/_app.js
@@ -6,13 +6,8 @@ import "highlight.js/styles/github.css";
 import "focus-visible/dist/focus-visible";
 import dynamic from "next/dynamic";
 import { QueryClient, QueryClientProvider } from "react-query";
-
-const HeadSEO = dynamic(() => import("../src/components/HeadSEO"), {
-  ssr: false,
-});
-const HeadLinks = dynamic(() => import("../src/components/HeadLinks"), {
-  ssr: false,
-});
+import HeadLinks from "../src/components/HeadLinks";
+import HeadSEO from "../src/components/HeadSEO";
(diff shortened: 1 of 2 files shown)


frontend/pages/_app.js
```diff
@@ -6,13 +6,8 @@ import "highlight.js/styles/github.css";
...
-
-const HeadSEO = dynamic(() => import("../src/components/HeadSEO"), {
-  ssr: false,
-});
-const HeadLinks = dynamic(() => import("../src/components/HeadLinks"), {
-  ssr: false,
-});
+import HeadLinks from "../src/components/HeadLinks";
+import HeadSEO from "../src/components/HeadSEO";
 const AppContext = dynamic(() => import("../src/AppContext"), {
```
zomglings, line 11: Does `AppContext` need to load dynamically?

peersky: @zomglings Build will fail if you try running it with normal import with following error:
`ReferenceError: window is not defined`

This happens because then it will try to render full HTML of those pages (bodies), while with dynamic it renders only `<head>` and leaves rest as `<script>`

peersky: Also rendering full HTML does not seem to be making sense for loading speed since when I was trying that - it appears that it will generate BOTH - HTML and same script that will run again after loading, hence chunk size is even bigger and HTML prerendered body indeed can be quite big

zomglings (approved):
Lgtm

PR #351, opened 2021-10-31
Abi upload with code editor block in modal

### LINKED ISSUE #344: New frontend component "ABI Uploader"
We want to be able user to upload ABI for smart contracts. For that user need to have UI component that can take ABI json file upload or take clipboard / text area input 

Component should be modal with state controlled trough `UIProvider` so it can be reused across multiple different pages

peersky:
## Changes

- in `/subscriptions` you can see `upload` buttons in `abi` column
- new modal type for uploading abi
- Verifying json

This is purely UI PR, no backend/logic implemented

## How to test these changes?

Locally see that UI is OK

## Related issues

resolves #344

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -25,6 +25,7 @@
     "next": "11.0.1",
     "nprogress": "^0.2.0",
     "react": "^17.0.2",
+    "react-ace": "^9.5.0",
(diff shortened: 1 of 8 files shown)


Andrei-Dolgolev (approved):
Looks correct.

PR #354, opened 2021-11-01
Add load older events in second request.

Andrei-Dolgolev:
## Changes

Add adition state to use effect on EntriesNavigator wich wait state when cache still empty list but we get `olderEvent.event_timestamp` have flag wich protect from execution that if state again maybe unrequired need check later.
Fix `settings.py` variable must be set error.


kompotkot (approved):
lgtm

PR #359, opened 2021-11-02
Imrovements for db go server

kompotkot:
## Changes

Fixed issue with CORS and better server arch.


Andrei-Dolgolev (approved):
Looks correct

PR #367, opened 2021-11-04
Removed unnecessary flag at service crawler

kompotkot:
## Changes

Fixed start of service for txpool crawler


Andrei-Dolgolev (approved):
Looks correct

PR #384, opened 2021-11-10
Setup instead of requirements.txt

kompotkot:
## Changes

In order to avoid deployment issues with pip and have one place to trach package, `requirements.txt` file has been deleted and all package requirements transfered to `setup.py`. Also added release workflows yml files.

`backend/moonstream` renamed to `backend/moonstreamapi` because moonstream name under client library

## How to test these changes?

Tested installations and main functions execution.


Andrei-Dolgolev (approved):
Looks correct

kompotkot:
@bugout-dev check
- [ ] Release after merge

PR #416, opened 2021-11-15
Prevent user-data script fall

kompotkot:
## Changes

During new node creation to prevent user-data script fall better to check file exist if not copy from S3


Andrei-Dolgolev (approved):
Looks correct

PR #438, opened 2021-11-16
using correct callback to close newDashboard drawer

peersky:
## Changes

this will correctly clear state when drawer cancel button is clicked in new dashboard creation


zomglings (approved):
lgtm

PR #439, opened 2021-11-16
use updated color keys in range selector styling

peersky:
## Changes

fix colors on range selector to work on active focus hover


zomglings (approved):
lgtm

PR #445, opened 2021-11-17
fix login and signup modal call from  sidebar

peersky:
## Changes

Fixed  mobile sidebar not opening modal signup and login


Yhtiyar (approved):
lg

PR #452, opened 2021-11-23
Add apply of filters.

Andrei-Dolgolev:
Add apply of filters wich in general depends on names of dashboard subscriptions.
Values filters not include.

Reformat code for get support of subscription wich not contain abi and only have set of generic methods.

Remove filtering labels by dates so if label vailabel in database and in filter/abi if no data for that labels query return zeros timesseries.


kompotkot (approved):
lgtm

PR #457, opened 2021-11-24
Client create stream

kompotkot:
## Changes

Now users with `moonstream` python client library are able to attach to stream of events.

Following chages:
- filtered events (if user required only `ethereum_blockchain` he will get it without other like whalewatch, txpool, etc)
- reversed events (if user provide start_time less then end time, he will get events in reversed order)

## How to test these changes?

Tested locally.


Andrei-Dolgolev (approved):
Looks correct

PR #460, opened 2021-11-24
Txpool working with both blockchains

kompotkot:
## Changes

Now you can chose on which blockchain to run crawler.


Andrei-Dolgolev (approved):
Looks correct

PR #466, opened 2021-11-25
fix query keys to be correct to stop refetches

peersky:
## Changes

Query keys were defined as randomly generated uuids, i changed those to be `dashboardId` and `timeRange`
This should make query key correspond to time range and dashboard as it should be. 

## How to test these changes?

Run and test that s3 objects is being pulled only once 

## Related issues
resolve https://github.com/bugout-dev/moonstream/issues/464


Andrei-Dolgolev (approved):
Looks correct

PR #468, opened 2021-11-26
Add timeout for load statistics with rerequest.

Andrei-Dolgolev:
## Changes
Apply timeout on request to S3 getFromPresignedURL but it's only response waiting timeout it's not catch connection timeout.

## Related issues
resolve https://github.com/bugout-dev/moonstream/issues/464

### DIFF
--- a/frontend/src/core/hooks/usePresignedURL.js
+++ b/frontend/src/core/hooks/usePresignedURL.js
@@ -19,7 +19,7 @@ const usePresignedURL = ({
       // You can uncomment this to use mockupsLibrary in development
       // url: `https://example.com/s3`,
       method: "GET",
-      timeout: 15 * 1000,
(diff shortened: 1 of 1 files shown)


frontend/src/core/hooks/usePresignedURL.js
```diff
@@ -29,12 +30,15 @@ const usePresignedURL = ({
     {
       ...queryCacheProps,
       enabled: isEnabled && url ? true : false,
+      retry: 3,
```
peersky, line 33: I think it is better to process retry as function 

```js
  retry: (failureCount, error) => {
    const status = error?.response?.status;
    return  status === 404 || status === 403 || (status === 408 && failureCount>3) ? false : true;
```

peersky (changes requested):

PR #479, opened 2021-12-04
Apply information about blocks to stats report.

Andrei-Dolgolev:
## Changes

Add next structure to dashboards stats:
```json
{
    "latest_available_block": 22119823,
    "latest_crawled_block": 22119756,
    "earliest_crawled_block": 11029941
}
```

## How to test these changes?
Generate stats dashboards. Polygon as example:
```bash
python -m mooncrawl.stats_worker.dashboard generate --blockchain polygon
```

### DIFF
--- a/crawlers/mooncrawl/mooncrawl/stats_worker/dashboard.py
+++ b/crawlers/mooncrawl/mooncrawl/stats_worker/dashboard.py
@@ -420,9 +420,9 @@ def get_blocks_state(
     """
 
     blocks_state = {
-        "latest_available_block": 0,
-        "latest_crawled_block": 0,
-        "earliest_crawled_block": 0,
+        "latest_stored_block": 0,
+        "latest_labelled_block": 0,
+        "earliest_labelled_block": 0,
     }
 
     label_model = get_label_model(blockchain_type)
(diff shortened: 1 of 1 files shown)


crawlers/mooncrawl/mooncrawl/stats_worker/dashboard.py
```diff
@@ -411,6 +411,60 @@ def get_unique_address(
     )
 
 
+def get_blocks_state(
+    db_session: Session, blockchain_type: AvailableBlockchainType, address: str
+):
```
zomglings, line 416: @Andrei-Dolgolev : Please put type annotation for return value.

crawlers/mooncrawl/mooncrawl/stats_worker/dashboard.py
```diff
@@ -411,6 +411,60 @@ def get_unique_address(
...
+
+    """
+    Generate meta information about
+    """
+
+    blocks_state = {
+        "blocks_state": {
+            "latest_block": {"address": 0, "labels": 0, "transactions": 0},
```
zomglings, line 424: This feels like too much nesting. Do we really need a `"blocks_state"` key? It is only needed inside the stats object.

Also, I would expect to just see something like this:
```
{
    "latest_available_block": 22119823,
    "latest_crawled_block": 22119756,
    "earliest_crawled_block": 11029941
}
```

Not sure what is value of `"labels"` and `"transactions"` and in general the schema you have seems too complicated.

crawlers/mooncrawl/mooncrawl/stats_worker/dashboard.py
```diff
@@ -411,6 +411,60 @@ def get_unique_address(
...
+    ).subquery(name="max_transactions_block")
+
+    result = (
+        db_session.query(
+            func.min(label_model.block_number),
+            func.max(label_model.block_number),
+            max_transactions_number.c.block_number,
+            max_labels_block.c.block_number,
+        )
+        .filter(label_model.address == address)
```
zomglings, line 448: I don't expect to see this filter. Otherwise you are making redundant functionality with `latest_events` on the provider side.

This method is just to get information about crawler status, not about particular events for this address.

crawlers/mooncrawl/mooncrawl/stats_worker/dashboard.py
```diff
@@ -685,6 +745,8 @@ def stats_generate_handler(args: argparse.Namespace):
 
                         print(f"Timescale: {timescale}")
 
+                        s3_data_object["blocks_state"] = current_blocks_state
```
zomglings, line 748: So now in the S3 data you have:
```
{
    ....,
    "blocks_state": {
        "blocks_state": {
            ....
        }
    },
    ....
}
```

Too complex. Make it simpler please.

zomglings (changes requested):
I think the schema for the state object is too complex and it is doing more than representing state. Please have a look at my comments.

Andrei-Dolgolev:
latest_available_block -> latest_stored_block 
latest_crawled_block-> latest_labelled_block

kompotkot (approved):
lgtm

PR #500, opened 2021-12-16
Fix - Polygon syncronize crawler works in multi threads

kompotkot:
## Changes

1 thread for polygon syncronization crawler is too slow.

## How to test these changes?

Already in prod


Yhtiyar (approved):
lg

PR #510, opened 2021-12-21
Continuous crawler batch saving

Yhtiyar:
## Changes
1. Crawler now will save labels once in `heartbeat_interval` (default 1 min)
2. Confirmations count increased to 175, so `moonworm-crawler` will be slower that blocks crawler and can use db more frequently


zomglings (approved):
lgtm

PR #515, opened 2021-12-22
Ethereum nodes deploy as polygon

kompotkot:
## Changes

Unified deployments for ethereum node as it works now for polygon, also simplified node status server to work with local ip

## How to test these changes?

Partly


Andrei-Dolgolev (approved):
Looks correct

kompotkot:
@bugout-dev check
- [ ] Deploy when new node will be ready at aws

PR #519, opened 2022-01-11
Node balancer re-use client hot node

kompotkot:
## Changes

In order to prevent crawlers jump from one node to another between one crawl, was added client structure which saves hot nodes for IP addr for short period of time.

## How to test these changes?

Tested with production nodes

## Related issues

Resolves https://github.com/bugout-dev/moonstream/issues/520


kompotkot:
@bugout-dev check

- [ ] First merge PR https://github.com/bugout-dev/moonstream/pull/515

nodes/node_balancer/cmd/clients.go
```diff
@@ -0,0 +1,97 @@
...
+
+import (
+	"log"
+	"reflect"
+	"time"
+
+	configs "github.com/bugout-dev/moonstream/nodes/node_balancer/configs"
+)
+
+var clientPool ClientPool
```
zomglings, line 11: I would prefer a pattern where you have:

```
func CreateClientPool() (*ClientPool, error) {
    ...
}
```

Then call `CreateClientPool` when you start the server and pass client pool reference wherever it is needed.

nodes/node_balancer/cmd/clients.go
```diff
@@ -0,0 +1,97 @@
...
+
+// Add client node and client itself if doesn't exist
+// TODO(kompotkot): Add mutes as for balancer
+func (cpool *ClientPool) AddClientNode(ip, blockchain string, node *Node) {
```
zomglings, line 15: instead of `ip`, use `clientID`. It will allow us to generalize in the future.

nodes/node_balancer/cmd/clients.go
```diff
@@ -0,0 +1,97 @@
...
+		client.ClientNodes = append(client.ClientNodes, ClientNode{
+			Blockchain: blockchain,
+			Node:       node,
+			LastCallTs: ts,
+		})
+	}
+}
+
+// Get client hot node if exists
+// TODO(kompotkot): Add mutes as for balancer
```
zomglings, line 52: mutes -> mutex?

nodes/node_balancer/cmd/data.go
```diff
@@ -17,6 +17,23 @@ type NodeStatusResponse struct {
...
+	LastCallTs int64
+}
+
+type Client struct {
+	IP          string
+	ClientNodes []ClientNode
+}
+
+type ClientPool struct {
+	Clients []*Client
```
zomglings, line 34: I would instead expect:

```
type ClientPool struct {
    Clients map[string]*Node
}
```

Do you have a use case for multiple client nodes per client?

Instead of putting blockchain in the `ClientNode` struct, I would recommend creating two client pools variables on the server - `ethereumClientPool` and `polygonClientPool`. I think that generalizes better.

zomglings (commented):
@kompotkot : A few comments.

nodes/node_balancer/cmd/clients.go
kompotkot: But I need to store this variable globally, still I will need to declare it same way..

nodes/node_balancer/cmd/data.go
kompotkot: Made this as map[string]*Node, but with different polygon and ethereum clientPools it brings a lot of code duplication. Because I am using structure:
```
func (cpool *ClientPool) AddClientNode(id, blockchain string, node *Node) {}
```
And can't inherit from `ethereumClientPool` and `polygonClientPool` same time, so it will require code duplication

nodes/node_balancer/cmd/clients.go
kompotkot: Can do this way

```
func CreateClientPool() {
	ethereumClientPool.Client = make(map[string]*Client)
	polygonClientPool.Client = make(map[string]*Client)
}
```

nodes/node_balancer/cmd/clients.go
kompotkot: Done, it reades client header

kompotkot:
@bugout-dev check
- [ ] Update web3 uri env vars

zomglings (approved):
lgtm

This downgrades performance of crawlers, so requires at least one round of optimization. Merging and deploying to drones server, but not setting crawlers to use node balancer yet.

PR #532, opened 2022-01-20
Heimdall now work via nodebalancer

kompotkot:
## Changes

Heimalld from polygon node works via node balancer instead of direct one node.

## How to test these changes?

Already in production


Yhtiyar (approved):
lg

PR #547, opened 2022-01-31
Dasboard update timer modified from 15m -> 6h

kompotkot:
## Changes

We update button there are no needs in dashboard frequently auto updates.

## How to test these changes?

No need tests


Andrei-Dolgolev (approved):
Looks correct

PR #555, opened 2022-02-22
fixed sleep time bug

Yhtiyar:
There was the bug in the logic of the `sleep_time` which was added to decrease the pressure to the node.
The bug was : `The sleep time was only increasing, but never decreasing, so sleep_time could become ~1-2 hours`
## Changes
removed exponential increase of the sleep time. added sleep time decrease.


Yhtiyar:
also increased `min_blocks_batch` from 10 to 40, so it will slightly decrease pressure on node.

PR #567, opened 2022-03-17
Crawlers access id to work with nodebalancer

kompotkot:
## Changes

To be able to connect to nodebalancer, `access_id` should be specified.

## How to test these changes?

Tested locally


kompotkot:
@bugout-dev check
- [ ] Add env variable `NB_CONTROLLER_ACCESS_ID`

crawlers/mooncrawl/mooncrawl/contract/cli.py
```diff
@@ -117,7 +119,7 @@ def run_crawler_desc(
 
 def handle_parser(args: argparse.Namespace):
     with yield_db_session_ctx() as session:
-        w3 = connect(AvailableBlockchainType.ETHEREUM)
+        w3 = connect(AvailableBlockchainType.ETHEREUM, access_id=args.access_id)
```
zomglings, line 122: @kompotkot : Does this work even if we want to connect directly to Ethereum node (skipping node balancer)?

zomglings (commented):
@kompotkot : Please confirm that setting access ID headers doesn't change behaviour of crawlers when using nodes directly (and not using node balancer).

We may need this, for example, if node balancer goes down and we have to fall back directly to crawling from nodes.

crawlers/mooncrawl/mooncrawl/contract/cli.py
kompotkot: Yes, if None provided by default it do not change behaviour.

PR #572, opened 2022-03-24
Moonstream db status server response with polygon block

kompotkot:
## Changes

DB status server also response with polygon block as for ethereum.

## How to test these changes?

Tested locally


zomglings (approved):
lgtm

PR #583, opened 2022-04-14
checking function abi before creating job

Yhtiyar:
## Changes
When function call crawler job is created, checking function signature if this job already exists

### DIFF
--- a/crawlers/mooncrawl/mooncrawl/moonworm_crawler/crawler.py
+++ b/crawlers/mooncrawl/mooncrawl/moonworm_crawler/crawler.py
@@ -154,12 +154,14 @@ def get_crawl_job_entries(
             offset=current_offset,
             limit=limit,
         )
+
         entries.extend(search_result.results)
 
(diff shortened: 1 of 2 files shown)


crawlers/mooncrawl/mooncrawl/moonworm_crawler/crawler.py
```diff
@@ -217,7 +219,15 @@ def make_function_call_crawl_jobs(
...
-            crawl_job_by_address[contract_address].contract_abi.append(json.loads(abi))
+            old_selectors = [
+                encode_function_signature(function_abi)
+                for function_abi in crawl_job_by_address[contract_address].contract_abi
+            ]
+            new_selector = encode_function_signature(json.loads(abi))
+            if new_selector not in old_selectors:
+                crawl_job_by_address[contract_address].contract_abi.append(
+                    json.loads(abi)
+                )
```
Yhtiyar, line 230: This is the most important part

zomglings (approved):
lgtm

PR #585, opened 2022-04-18
fix v2.

Yhtiyar:


zomglings (approved):
lgtm

Perhaps we should wait to deploy this until after @Andrei-Dolgolev has set stats generator to use the replica DB.

PR #599, opened 2022-05-05
Add migration for create index  for speed up queries.

Andrei-Dolgolev:
Add migration for create index in (address, label, (label_data->>'type'), (label_data->>'name')) in polygon labels.


kompotkot (approved):
lg

PR #613, opened 2022-05-23
xdai db model

kompotkot:
## Changes

DB model and alembic migration of XDai blockchain

## How to test these changes?

Tested locally


zomglings (approved):
gg

PR #619, opened 2022-05-26
XDai database blockchain update to support XDai

kompotkot:
## Changes

Blockchain XDai extension.


Andrei-Dolgolev (approved):
Looks correct

PR #622, opened 2022-05-26
Using replica read only to get subscription events

kompotkot:
## Changes

Expected speed will be increased.


Andrei-Dolgolev (approved):
Looks correct

PR #623, opened 2022-05-26
Revert "Using replica read only to get subscription events"

kompotkot:
Reverts bugout-dev/moonstream#622


Andrei-Dolgolev (approved):
Looks correct

PR #634, opened 2022-06-22
Moonworm historical crawler

Yhtiyar:
## Changes
Added historical crawler
Example:
```bash
moonworm-crawler historical-crawl --address 0x431CD3C9AC9Fc73644BF68bF5691f4B83F9E104f --poa --end 25329797 --max-blocks-batch 5000 --only-events -b polygon
```


Andrei-Dolgolev (approved):
Looks correct

PR #636, opened 2022-07-11
Sourcing env file with nix start and new commit with fix

kompotkot:
## Changes

Small changes for scm db

## How to test these changes?

tested on prod


Andrei-Dolgolev (approved):
Looks correct

PR #637, opened 2022-07-11
External nodes support

kompotkot:


kompotkot:
@bugout-dev check
- [ ] Modify config file according to new json structure

kompotkot:
@bugout-dev check accept Modify config file according to new json structure

PR #640, opened 2022-07-15
Callcounter reset when 10mln

kompotkot:
## Changes

To prevent long future overflow.

## How to test these changes?

Tested locally


zomglings (approved):
gg

(Didn't review properly)

PR #653, opened 2022-08-22
Changing logo on landing page.

kellan-simiotics:


kellan-simiotics:
> gg
> 
> But on large screens, the logo is further from the left side of the screen than the final text is from the right side of the screen. Might want to review margins @kellan-simiotics , @ishihssihshihishsi .

Done.

PR #660, opened 2022-08-31
Metadata crawler

Andrei-Dolgolev:


crawlers/mooncrawl/mooncrawl/metadata_crawler/cli.py
```diff
@@ -0,0 +1,163 @@
...
+
+    """
+    Parse all metadata of tokens.
+    """
+
+    logger.info("Starting metadata crawler")
+    logger.info(f"Connecting to blockchain {blockchain_type.value}")
+
+    engine = create_moonstream_engine(
+        MOONSTREAM_DB_URI_READ_ONLY,
```
kompotkot, line 70: not read only if we are writing

kompotkot (changes requested):
Small fixes, also can we have a description what this crawler do, otherwise nobody understand it.

kompotkot (approved):
lg

PR #665, opened 2022-09-12
State crawler improvments

Andrei-Dolgolev:

### DIFF
--- a/crawlers/mooncrawl/mooncrawl/metadata_crawler/cli.py
+++ b/crawlers/mooncrawl/mooncrawl/metadata_crawler/cli.py
@@ -4,6 +4,7 @@ from urllib.error import HTTPError
 import urllib.request
 import logging
 from typing import Dict, Any
+from concurrent.futures import ThreadPoolExecutor
 
 from moonstreamdb.blockchain import AvailableBlockchainType
 from moonstreamdb.db import (
@@ -19,8 +20,10 @@ from .db import (
     metadata_to_label,
 )
 from ..settings import (
+    MOONSTREAM_METADATA_CRAWLER_THREADS,
     MOONSTREAM_STATE_CRAWLER_DB_STATEMENT_TIMEOUT_MILLIS,
 )
+from ..data import TokenURIs
 
 logging.basicConfig(level=logging.INFO)
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


crawlers/mooncrawl/mooncrawl/metadata_crawler/cli.py
```diff
@@ -38,8 +44,14 @@ def crawl_uri(metadata_uri: str) -> Any:
     result = None
     while retry < 3:
```
kompotkot, line 45: I sugest to re-write this func to avoid increment of `retry` var only in exceptions. But would do this by default at the end of loop and `if response.status == 200:` then write result and break the `while` loop.
But if this version works well, then probably there are no need any changes.

crawlers/mooncrawl/mooncrawl/metadata_crawler/cli.py
```diff
@@ -78,39 +95,56 @@ def parse_metadata(blockchain_type: AvailableBlockchainType, batch_size: int):
...
+                address=address,
+            )
 
-            already_parsed = get_current_metadata_for_address(
-                db_session=db_session, blockchain_type=blockchain_type, address=address
+            logger.info(
+                f"Start crawling {len(not_updated_tokens)} tokens of address {address}"
             )
 
             for requests_chunk in [
```
kompotkot, line 112: Hard to read, probably better to add another variable on top of this `for` loop.

crawlers/mooncrawl/mooncrawl/metadata_crawler/cli.py
```diff
@@ -78,39 +95,56 @@ def parse_metadata(blockchain_type: AvailableBlockchainType, batch_size: int):
...
-                for i in range(0, len(tokens_uri_by_address[address]), batch_size)
+                not_updated_tokens[i : i + batch_size]
+                for i in range(0, len(not_updated_tokens), batch_size)
             ]:
                 writed_labels = 0
-                for token_uri_data in requests_chunk:
 
-                    if token_uri_data.token_id not in already_parsed:
-                        metadata = crawl_uri(token_uri_data.token_uri)
+                with ThreadPoolExecutor(
```
kompotkot, line 118: With thread pool executor better to store results and `wait` them like in this code https://github.com/bugout-dev/moonstream/blob/e5240d2fd85685383bb40ea4cad741ec90b4fd6d/crawlers/mooncrawl/mooncrawl/blockchain.py#L367

kompotkot: So this threads which inside another loop from list comprehension range will be under better control))

kompotkot (commented):
some comments

PR #669, opened 2022-09-14
Add remove worker

Andrei-Dolgolev:
## How to test these changes?

```bash 
state-crawler clean-state-labels -b polygon -N 10000
```


kompotkot (approved):
lg

PR #670, opened 2022-09-15
Different timer on boot start time

kompotkot:
## Changes

Changes to avoid statement timeout response from database.


Andrei-Dolgolev (approved):
looks correct

PR #680, opened 2022-10-15
Added transaction rollback when there is a violation of unique constraint when crawling blocks

zomglings:
## Changes

This was causing an error in production.


kompotkot (approved):
lg

PR #686, opened 2022-10-25
Query crawler

kompotkot:
## Changes

Fetch query from journal, validate query, execute query and push to bucket if required.

## How to test these changes?

Tested locally

### DIFF
--- a/crawlers/mooncrawl/mooncrawl/queries_crawler/cli.py
+++ b/crawlers/mooncrawl/mooncrawl/queries_crawler/cli.py
@@ -39,14 +39,17 @@ def parser_queries_execute_handler(args: argparse.Namespace) -> None:
             data_rows=data_rows,
         )
 
-        if args.output is not None and args.upload:
-            bucket_metadata = {"source": "queries-crawler"}
-            push_data_to_bucket(
-                data=output,
-                key=f"{MOONSTREAM_S3_DATA_BUCKET_PREFIX}/queries/{str(args.id)}/data.{args.output}",
-                bucket=MOONSTREAM_S3_DATA_BUCKET,
-                metadata=bucket_metadata,
-            )
+        if args.output is not None:
+            if args.upload:
+                bucket_metadata = {"source": "queries-crawler"}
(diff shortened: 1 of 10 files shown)


kompotkot:
@bugout-dev check
- [ ] Add env MOONSTREAM_QUERIES_JOURNAL_ID
- [ ] Add env MOONSTREAM_S3_DATA_BUCKET
- [ ] Add env MOONSTREAM_S3_DATA_BUCKET_PREFIX

crawlers/mooncrawl/mooncrawl/queries_crawler/cli.py
```diff
@@ -0,0 +1,91 @@
...
+    fetch_query_from_journal,
+    prepare_output,
+    query_validation,
+)
+
+logging.basicConfig(level=logging.INFO)
+logger = logging.getLogger(__name__)
+
+
+def parser_queries_execute_handler(args: argparse.Namespace) -> None:
```
zomglings, line 20: After discussing with @kompotkot and @Andrei-Dolgolev:

We realized that we want this crawler to go through the Query API like any other user would.

The plan is to remove this `mooncrawl/queries_crawler` code and update the Moonstream Python client to provide this functionality similarly to how we used it in `autocorns biologist`:
https://github.com/bugout-dev/autocorns/blob/cf00fb492de254821730a256d238d5a332810db6/autocorns/biologist.py#L371

We can remove the existing Moonstream Python client, bump the client version, and publish the new client.

zomglings (commented):
Reconsidered crawler architecture.

kompotkot:
We need to cherry pick from this PR later.

zomglings:
Although we will not use `queries_crawler` to crawl public data, we *will* use it as a replacement for the existing Query API data producer (which is currently in `mooncrawl/stats_worker/queries.py`.

The `queries_crawler` should be a CLI which exposes the same functionality but in a more modular way (and should be invoked from systemd on prod).

We will revisit this after our current batch of urgent work.

PR #687, opened 2022-10-26
New python client query api based

kompotkot:
## Changes

Support of query API with moonstream python library.

## How to test these changes?

Tested locally


Andrei-Dolgolev (approved):
Looks correct

PR #689, opened 2022-11-01
X-Node-URL response header on node balancer responses

zomglings:
## Changes

Proposed addition of X-Node-URL to response header on node balancer responses.


zomglings:
Stale. Will update relative to latest nodebalancer changes.

PR #692, opened 2022-11-03
Fix release client library github action

kompotkot:
## Changes

Fixed client release workflow action.


Andrei-Dolgolev (approved):
Looks correct

PR #695, opened 2022-11-09
Deployment upgrades to work with systems

kompotkot:
## Changes

- Migrated deployment to checkenv
- Backend working under regular user


Andrei-Dolgolev (approved):
Looks correct

PR #698, opened 2022-11-09
Fix text color

Anton-Mushnin:
## Changes
Changing Contact Form background color to make text more visible.
Changing text color on Streams Filter Drawer

also changing logo cursor in sidebar to pointer


kellan-simiotics (approved):
Looks good.

PR #715, opened 2022-11-24
Add parameters in s3 path

Andrei-Dolgolev:
## Changes
Main:
Previous S3 path keys structure.
```
Bucket: MOONSTREAM_S3_QUERIES_BUCKET
Key: {MOONSTREAM_S3_QUERIES_BUCKET_PREFIX}/queries/{query_id}/data.{file_type}
```
was change to

```
Bucket: 
MOONSTREAM_S3_QUERIES_BUCKET
Key: {MOONSTREAM_S3_QUERIES_BUCKET_PREFIX}/queries/{query_id}/{hash_of_passed_parameters}/data.{file_type}
```
It will broke logic of `/get_links` endpoint. Because for restore access to s3 route require pass query parameters payload for repeat hash generation. 

Fixes:
1) One of way is add reading params from body in GET request which not disallowed but not recommended.
2) @zomglings suggest pass parameters say `update=false` query parameter and then drones return only presign_url without requesting update data on S3 and delete GET query_links endpoint complitly.

Testing via usual updated request:
```
POST https://api.moonstream.to/queries/{query_name}/update_data
Content-Type: application/json
authorization: Bearer <token>
```

The drones server must return new version of s3 path.


crawlers/mooncrawl/mooncrawl/stats_worker/queries.py
```diff
@@ -86,27 +83,36 @@ def data_generate(
...
+        "file_type": file_type,
+        "params_hash": params_hash,
+        "params": json.dumps(params),
+    }
+
     try:
+
+        # TODO:(Andrey) Need optimization that information is usefull but incomplete
+        block_number, block_timestamp = db_session.execute(
+            "SELECT block_number, block_timestamp FROM polygon_labels WHERE block_number=(SELECT max(block_number) FROM polygon_labels where label='moonworm-alpha') limit 1;",
```
zomglings, line 98: What if it is an Ethereum query?

crawlers/mooncrawl/mooncrawl/stats_worker/queries.py
```diff
@@ -118,12 +124,12 @@ def data_generate(
...
             ).encode("utf-8")
-            push_statistics(
-                s3=s3,
-                data=data,
-                key=f"{MOONSTREAM_S3_QUERIES_BUCKET_PREFIX}/queries/{query_id}/data.{file_type}",
-                bucket=bucket,
-            )
+        push_data_to_bucket(
+            data=data,
+            key=key,
```
zomglings, line 129: Where is this key generated?

crawlers/mooncrawl/mooncrawl/api.py
```diff
@@ -201,40 +227,36 @@ async def queries_data_update_handler(
...
-        valid_query = queries.query_validation(request_data.query)
-    except queries.QueryNotValid:
-        logger.error(f"Incorrect query provided with id: {query_id}")
-        raise MoonstreamHTTPException(
-            status_code=401, detail="Incorrect query provided"
-        )
-    except Exception as e:
-        logger.error(f"Unhandled query execute exception, error: {e}")
-        raise MoonstreamHTTPException(status_code=500)
+    params_hash = hashlib.md5(
```
zomglings, line 230: It makes sense to make this its own function - like `def query_parameter_hash(passed_params): ...`.

zomglings (commented):
Some questions.

crawlers/mooncrawl/mooncrawl/stats_worker/queries.py
Andrei-Dolgolev: Rigth now in question TODO exactly about it)

crawlers/mooncrawl/mooncrawl/stats_worker/queries.py
Andrei-Dolgolev: Any Api call level once for presignUrl and bucket push

crawlers/mooncrawl/mooncrawl/api.py
Andrei-Dolgolev: Yes will move it

kompotkot (approved):
lg, only versions update required

Issue #717, opened 2022-11-28
No more hung rollbacks!


zomglings:
We are seeing too many production errors on Moonstream crawlers and API that look like this:

```
sqlalchemy.exc.PendingRollbackError: Can't reconnect until invalid transaction is rolled back. (Background on this error at: https://sqlalche.me/e/14/8s2b)
```

([SQLAlchemy documentation on error details](https://docs.sqlalchemy.org/en/14/errors.html#error-8s2b))

We should go through the code, `git grep ".commit("` and make sure that we are correctly handling exceptions and rolling back transactions *everywhere*.

Andrei-Dolgolev:
closed https://github.com/bugout-dev/moonstream/pull/727

PR #727, opened 2022-12-05
Add rollback.

Andrei-Dolgolev:


Andrei-Dolgolev:
Resolve https://github.com/bugout-dev/moonstream/issues/717

PR #735, opened 2022-12-14
Fix deprecated keys for gnosis

kompotkot:
## Changes

According with merge json-rpc request removed unused keys for gnosis block.


Andrei-Dolgolev (approved):
Looks correct

PR #741, opened 2022-12-19
Supports tags filter for nodes

kompotkot:
## Changes

Nobebalancer support of filter by tag.

## How to test these changes?

Tested locally

PR #747, opened 2023-01-12
Add x-node-url response header to requests

zomglings:
## Changes

The `x-node-url` response header specifies which node a JSONRPC request served by a blockchain (not database) was served from.

## How to test these changes?

Tested locally with curl:

```
curl -v --request POST "http://127.0.0.1:8544/nb/polygon/jsonrpc?access_id=$NB_CONTROLLER_ACCESS_ID&data_source=blockchain" \
    --header 'Content-Type: application/json' \
    --data-raw '{
        "jsonrpc":"2.0",
        "method":"eth_blockNumber",
        "id":1
    }'
```


zomglings:
@bugout-dev check
- [ ] Depends on https://github.com/bugout-dev/moonstream/pull/741

zomglings:
I don't want to merge this. I would like to add a `name` field for nodes and then use the name in a response header instead of hte full JSONRPC URL. The current implementation could leak access keys to third party node providers.

PR #751, opened 2023-01-26
Add changes to subscription API for works with entity as subscription…

Andrei-Dolgolev:
…s data storage instead of brood resources.

### DIFF
--- a/backend/moonstreamapi/actions.py
+++ b/backend/moonstreamapi/actions.py
@@ -16,6 +16,8 @@ from bugout.data import (
 )
 from bugout.journal import SearchOrder
 from bugout.exceptions import BugoutResponseException
+from entity.data import EntityCollectionsResponse, EntityCollectionResponse
+from entity.exceptions import EntityUnexpectedResponse
 from ens.utils import is_valid_ens_name  # type: ignore
 from eth_utils.address import is_address  # type: ignore
 from moonstreamdb.models import EthereumLabel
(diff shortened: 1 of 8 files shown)


backend/moonstreamapi/admin/migrations/generate_entity_subscriptions.py
```diff
@@ -0,0 +1,651 @@
...
+    BUGOUT_RESOURCE_TYPE_DASHBOARD,
+)
+from ...settings import bugout_client as bc, entity_client as ec
+from ..subscription_types import CANONICAL_SUBSCRIPTION_TYPES
+
+logger = logging.getLogger(__name__)
+
+
+# Dashboard resource type
+BUGOUT_RESOURCE_TYPE_DASHBOARD = "dashboards"
```
kompotkot, line 34: Duplication of `..settings/BUGOUT_RESOURCE_TYPE_DASHBOARD`

backend/moonstreamapi/admin/migrations/generate_entity_subscriptions.py
```diff
@@ -0,0 +1,651 @@
...
+                        )
+                        stages[user_id]["subscription_resource_id"] = str(
+                            subscription_resource.id
+                        )
+                    except Exception as e:
+                        logger.error(
+                            f"Failed to create subscription brood resource: {str(e)}"
+                        )
+
+            if "proccessed_subscriptions" not in stages[user_id]:
```
kompotkot, line 298: Do we have typo `procCessed` in database, or fix here to `processed`

backend/moonstreamapi/admin/migrations/generate_entity_subscriptions.py
```diff
@@ -0,0 +1,651 @@
...
+                ] = {"entity_id": str(entity.entity_id), "dashboard_ids": []}
+
+            # Add permissions to user
+
+            if user_id != admin_user_id:
+                # Add permissions to user
+
+                if "permissions_granted" not in stages[user_id]:
+                    try:
+                        add_collection_permissions_to_user(
```
kompotkot, line 350: We are using this func only in one place, here, probably better to simplify code and remove function `add_collection_permissions_to_user` copy code from it here.

kompotkot (changes requested):
sm changes

kompotkot (commented):
comm

kompotkot (approved):
gg

PR #766, opened 2023-03-09
Custom engine for all crawlers with specific timeout

kompotkot:
## Changes

Some crawlers use short backend timeout, which is wrong and not let to finish long queries.


Andrei-Dolgolev (approved):
Looks correct

PR #792, opened 2023-05-31
Manual brood client creation with timeout

kompotkot:
## Changes

Manual brood client creation with timeout

## How to test these changes?

Tested locally


Andrei-Dolgolev (approved):
looks correct

PR #813, opened 2023-06-08
Workers with call_requests clean action

kompotkot:
Resolved Issuer https://github.com/moonstream-to/web3/issues/297


probs/README.md
```diff
@@ -0,0 +1,32 @@
+# probs
```
zomglings, line 1: `probs` -> `probes`

probs/cmd/probs/configs.go
```diff
@@ -0,0 +1,163 @@
...
+		}
+
+		var serviceWorkers []probs.ServiceWorker
+
+		// Link worker function
+		for w, worker := range service.Workers {
+			switch service.Name {
+			case "engine":
+				for _, engineWorker := range engine.ENGINE_SUPPORTED_WORKERS {
+					if worker.Name == engineWorker.Name {
```
zomglings, line 55: Not really an issue, but it would be a good idea to define a `map[string]Worker` in engine instead of iterating like this.

probs/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+Run service with configuration:
+
+```bash
+probs service --config "~/.probs/config.json"
+```
+
+Config example:
+
+```json
```
zomglings, line 19: Maybe better if this is not an array of services but rather a single service.

It is more consistent with the `probs service` command, and also less nesting in configuration is better.

zomglings (changes requested):
A few small comments. Let us deploy with `probes` as the name, and please consider my comment about unnesting the service definitions so that each JSON file represents a single service. Is there a use case for multiple services in the same file/process?

The architecture of this is very elegant, I like this new setup.

Andrei-Dolgolev (approved):
Looks correct

PR #815, opened 2023-06-12
Fix historical function crawl

Andrei-Dolgolev:
Fix incorrect using of entries object for function call.
And fix refetch(merge jobs logic) for events.


zomglings (approved):
lgtm

PR #820, opened 2023-06-16
Queries template

Andrei-Dolgolev:
## Add queries template logic over queries endpoints:
POST `/<query_name>/update_data`
POST `/<query_name>`

### **Template** 
Is bugout journal entry with context-type:`MOONSTREAM_QUERIES_TEMPLATE_CONTEXT_TYPE`
based on idea of @zomglings but without user resolution. 

template contain set of tags:
**query_template** - default tag mostly for simplify work with templates from bugout interface 

**query_url**:template_ownership_erc721 - e.g. query name is url of endpoint so that query will trigered by that query:
```
POST http://127.0.0.1:7481/queries/template_ownership_erc721/update_data
Authorization: Bearer <user_token>
Content-Type: application/json

{
    "blockchain": "polygon",
    "params": {
        "address": "0xdC0479CC5BbA033B3e7De9F178607150B3AbCe1f"
    }
}
```
Original idea was use context_id but for some reason context_id all time none for get entry endpoint until not recognize problem will be use tags instead.

**query_id**:742afd86-bba8-4a70-bb9d-e16cac94a668 - entry_id

**interface**:721 - split quries by interface supported if interfaces tags is present

**approved** - responsible for appear query in templates endpoints. So of it not present then template not visible for frontend.

 

### Add optional parameter `blockchain` 
in 
POST `/<query_name>/update_data` 

If that parameter present:
then before execution all keys:
`__transactions_table__`, `__blocks_table__`, `__labels_table__`
will resolve to table of coresponding blockchain.

So we escape problem with manage a lot of dublicated queries for different blockchain.

### Template creating

Added queries CLI in moonstream admin tool:
```
 mnstr queries create-template --query-file ./query.txt -n ownership-erc721
```
accept 2 parameters file name and query name.


kompotkot (approved):
lg

PR #821, opened 2023-06-17
Modified CORS middleware with Redis cache workflow

kompotkot:
At application startup:

- Fetch CORS settings from application resources application-config
- Fetch CORS from env variable and asure all this origins represented in resource application-config
- Write it to Redis cache

With `/cors` endpoint each user can add more origins to CORS application-config, after it the cache and 'allowed_origins' will be extended. List of all allowed origins available only to Moonstream admin.

As a source used design doc: https://docs.google.com/document/d/1ZGXAvGTH1zj3jr4hUWN-0-9VHrHcAgcQBYvywqg3WpM/edit#heading=h.5gdp4oo0iu7i

Tested locally.

### DIFF
--- a/engineapi/engineapi/api.py
+++ b/engineapi/engineapi/api.py
@@ -5,17 +5,15 @@ import logging
 import time
 
 from fastapi import FastAPI
-from fastapi.middleware.cors import CORSMiddleware
 
 from . import data
-from .settings import (
-    ORIGINS,
-)
+from .middleware import BugoutCORSMiddleware
+from .routes.admin import app as admin_app
+from .routes.configs import app as configs_app
 from .routes.dropper import app as dropper_app
 from .routes.leaderboard import app as leaderboard_app
-from .routes.admin import app as admin_app
-from .routes.play import app as play_app
 from .routes.metatx import app as metatx_app
+from .routes.play import app as play_app
 from .version import VERSION
 
 logging.basicConfig(level=logging.INFO)
@@ -34,8 +32,7 @@ app = FastAPI(
 )
 
 app.add_middleware(
-    CORSMiddleware,
-    allow_origins=ORIGINS,
+    BugoutCORSMiddleware,
     allow_credentials=True,
     allow_methods=["*"],
     allow_headers=["*"],
@@ -58,8 +55,9 @@ async def now_handler() -> data.NowResponse:
     return data.NowResponse(epoch_time=time.time())
 
 
+app.mount("/admin", admin_app)
+app.mount("/configs", configs_app)
 app.mount("/leaderboard", leaderboard_app)
 app.mount("/drops", dropper_app)
-app.mount("/admin", admin_app)
 app.mount("/play", play_app)
 app.mount("/metatx", metatx_app)
--- a/engineapi/engineapi/middleware.py
+++ b/engineapi/engineapi/middleware.py
@@ -10,7 +10,7 @@ from pydantic import AnyHttpUrl, parse_obj_as
 from starlette.middleware.base import BaseHTTPMiddleware
 from starlette.middleware.cors import CORSMiddleware
 from starlette.responses import Response
-from starlette.types import ASGIApp, Receive, Scope, Send
+from starlette.types import ASGIApp
 from web3 import Web3
 
 from .auth import (
@@ -18,7 +18,7 @@ from .auth import (
     MoonstreamAuthorizationVerificationError,
     verify,
 )
-from .rc import rc_client, yield_rc_async_session
+from .rc import REDIS_CONFIG_CORS_KEY, rc_client
 from .settings import (
     ALLOW_ORIGINS,
     BUGOUT_REQUEST_TIMEOUT_SECONDS,
@@ -299,8 +299,7 @@ def fetch_application_settings_cors_origins(token: str) -> Set[str]:
 
 def set_cors_origins_cache(allow_origins: Set[str]) -> None:
     try:
-        allow_origins_str = ",".join(list(allow_origins))
-        rc_client.set("cors", allow_origins_str)
+        rc_client.sadd(REDIS_CONFIG_CORS_KEY, *allow_origins)
     except Exception:
         logger.warning("Unable to set CORS origins at Redis cache")
     finally:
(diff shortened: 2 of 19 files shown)


engineapi/engineapi/api.py
```diff
@@ -58,8 +55,9 @@ async def now_handler() -> data.NowResponse:
     return data.NowResponse(epoch_time=time.time())
 
 
+app.mount("/admin", admin_app)
+app.mount("/configs", configs_app)
```
zomglings, line 59: Is `/configs` supposed to be for CORS only or also for other configuration?

zomglings: Is it for user-controlled configuration?

engineapi/engineapi/middleware.py
```diff
@@ -199,3 +212,157 @@ async def dispatch(
         request.state.token = user_token
 
         return await call_next(request)
+
+
+def parse_origins_from_resources(origins: List[str]) -> Set[str]:
```
zomglings, line 217: Does the resources here refer to Brood resources?

engineapi/engineapi/middleware.py
```diff
@@ -199,3 +212,157 @@ async def dispatch(
...
+    """
+    Parse list of CORS origins with HTTP validation and remove duplications.
+    """
+    resource_origins_set = set()
+    for resource in origins:
+        origins = resource.resource_data.get("origins", [])
```
zomglings, line 223: I really don't understand what this line of code is doing.

`origins` is the input variable, of type `List[str]`. `resource` is an item from `origins`, which means it is a string. What is `resource.resource_data`, then?

And why are you modifying the `origins` argument to the function inside the loop?

engineapi/engineapi/middleware.py
```diff
@@ -199,3 +212,157 @@ async def dispatch(
...
+
+    resource_origins_set: Set[str] = parse_origins_from_resources(resources.resources)
+    resource_origins_set = check_default_origins(resource_origins_set)
+
+    return list(resource_origins_set)
+
+
+def set_cors_origins_cache(allow_origins: Set[str]) -> None:
+    try:
+        allow_origins_str = ",".join(list(allow_origins))
```
zomglings, line 302: Redis is a key-value store. Why aren't we storing the actual origins as keys? It doesn't make sense to me that we are just joining all the origins into a comma-separated string and storing under the `cors` key.

What will happen if there are 1,000,000 origins that we need to whitelist?

I would use a Redis set to store the whitelisted origins: https://redis.io/docs/data-types/sets/

engineapi/engineapi/middleware.py
```diff
@@ -199,3 +212,157 @@ async def dispatch(
...
+            allow_origins=self.allow_origins,
+            allow_methods=allow_methods,
+            allow_headers=allow_headers,
+            allow_credentials=allow_credentials,
+            allow_origin_regex=None,
+            expose_headers=expose_headers,
+            max_age=max_age,
+        )
+
+    async def __call__(self, scope: Scope, receive: Receive, send: Send):
```
zomglings, line 347: I suggest:
1. If Redis connection is available, just check `SISMEMBER` on each request
2. If Redis connection is unavailable, check membership in application settings allowed origins on each request.

Why do we need to keep an `allow_origins` member around like this?

zomglings: We can read the hostname header on the request directly here and resolve CORS responses the same way that the original `CORSMiddleware` does. We don't need to maintain a separate `allow_origins` list if we can do a set membership check directly in Redis.

engineapi/engineapi/routes/configs.py
```diff
@@ -0,0 +1,161 @@
...
+
+app.add_middleware(
+    BugoutCORSMiddleware,
+    allow_credentials=True,
+    allow_methods=["*"],
+    allow_headers=["*"],
+)
+
+
+@app.get("/cors", response_model=data.CORSResponse)
```
zomglings, line 71: I think it's better if this endpoint is of the form: `/cors?hostname=<URL encoded hostname>`

It would basically return true if the hostname is in the CORS whitelist and false otherwise.

engineapi/engineapi/routes/configs.py
```diff
@@ -0,0 +1,161 @@
...
+            resources.resources
+        )
+    except Exception as err:
+        logger.error(repr(err))
+        raise EngineHTTPException(status_code=500)
+
+    return data.CORSResponse(cors=",".join(list(resource_origins_set)))
+
+
+@app.put("/cors", response_model=data.CORSResponse)
```
zomglings, line 95: Is this a public endpoint? Are we not checking user credentials before adding CORS?

zomglings (dismissed):
@kompotkot : I think this PR needs some work before we can merge it.

Please review my comments.

zomglings:
@kompotkot : Did you profile how long it takes to resolve requests using this new middleware compared to how long it took to resolve requests in previous scheme?

engineapi/engineapi/middleware.py
kompotkot: Yes

engineapi/engineapi/middleware.py
kompotkot: This resource is where we store user provided origins. So it could have multiple of them. I decided to store this as one resource. Of course we can create for each origin new resource.

engineapi/engineapi/middleware.py
kompotkot: Yes, I will use it, I was just following old approach where we were storing CORS origins as string in environment variable.

engineapi/engineapi/middleware.py
kompotkot: We need to update `self.allow_origins` or totally rewrite this middleware class logic, what I am trying to avoid.

engineapi/engineapi/routes/configs.py
kompotkot: User will see only his origins he added before. Do you suggest to add additional endpoint otherwise frontend should have a way to fetch list of all origins user added.

engineapi/engineapi/routes/configs.py
kompotkot: We are using `token=request.state.token` in this endpoint.

kompotkot:
> @kompotkot : Did you profile how long it takes to resolve requests using this new middleware compared to how long it took to resolve requests in previous scheme?

No I did not run profile tests.

engineapi/engineapi/middleware.py
kompotkot: Remove modification of `origins`

engineapi/engineapi/routes/configs.py
kompotkot: Add as additional route

kompotkot:
@bugout-dev
- [ ] First add Redis to playbook

kompotkot:
@bugout-dev check
- [ ] Add env vars for redis

Andrei-Dolgolev (approved):
Looks correct

PR #831, opened 2023-07-04
Biling cli

Andrei-Dolgolev:
mnstr usage generate --month '2024-02' --user-id <user_id> --contracts '{"polygon": [<address_for_events_count>]}'  --output ./reports/customer-2024-02.json


moonstreamapi/moonstreamapi/admin/billing.py
```diff
@@ -0,0 +1,216 @@
...
+
+        repeat += 1
+
+        if repeat > max_retries:
+            logger.info("Too many retries")
+            break
+    return data_response.json()
+
+
+def collect_billing(
```
zomglings, line 91: Let us call this `collect_billing_information`. It doesn't actually calculate a dollar value, so `collect_billing` is a bit misleading.

moonstreamapi/moonstreamapi/admin/billing.py
```diff
@@ -0,0 +1,216 @@
...
+) -> Dict[str, Any]:
+    """
+    Collect billing information for a user.
+
+    By user_id or token.
+
+    Collected info:
+
+    Resources:
+        - queries
```
zomglings, line 105: We don't bill by query.

moonstreamapi/moonstreamapi/admin/billing.py
zomglings: Maybe better to call it `collect_usage_information`. For things like number of queries, that's a very important statistic for us to track even though we don't use it for billing.

moonstreamapi/moonstreamapi/admin/cli.py
```diff
@@ -505,6 +519,55 @@ def main() -> None:
         "-n", "--name", required=True, help="Name for the new query"
     )
     create_query_parser.set_defaults(func=queries.create_query_template)
+
+    billing_parser = subcommands.add_parser(
+        "billing", description="Manage Moonstream billing"
```
zomglings, line 524: `"billing"` -> `"usage"`?

moonstreamapi/moonstreamapi/admin/cli.py
```diff
@@ -505,6 +519,55 @@ def main() -> None:
...
+
+    generate_billing_parser.add_argument(
+        "--month",
+        required=True,
+        type=str,
+        help="Month for which to generate billing in YYYY-MM format (e.g. 2021-10)",
+    )
+
+    generate_billing_parser.add_argument(
+        "--user-token",
```
zomglings, line 543: Better if the input is `--user-id`, not `--user-token`.

zomglings (commented):
Suggested renaming `billing` to `usage`.

Also suggested `--user-id`, not `--user-token`.

Andrei-Dolgolev:
Block by issues

kompotkot (approved):
lg

PR #838, opened 2023-07-10
Set dependencies in requirements.txt for moonstreamapi

kompotkot:
Otherwise pydantic will throw an error.


Andrei-Dolgolev (approved):
Looks good

PR #882, opened 2023-08-03
Reorg metatx db

kompotkot:
Some database modifications for indexes


engineapi/engineapi/models.py
```diff
@@ -155,14 +157,87 @@ class DropperClaimant(Base):  # type: ignore
...
+
+    id = Column(
+        UUID(as_uuid=True),
+        primary_key=True,
+        default=uuid.uuid4,
+        unique=True,
+    )
+    name = Column(VARCHAR(128), nullable=False, index=True, unique=True)
+    chain_id = Column(Integer, nullable=False, index=True, unique=False)
+    testnet = Column(Boolean, default=False, nullable=False)
```
zomglings, line 225: Good idea.

engineapi/alembic/versions/b4257b10daaf_call_request_types_and_metatx_holders.py
```diff
@@ -0,0 +1,171 @@
...
+    )
+
+    op.create_index(op.f('ix_blockchains_chain_id'), 'blockchains', ['chain_id'], unique=False)
+    op.create_index(op.f('ix_blockchains_name'), 'blockchains', ['name'], unique=True)
+    
+    op.add_column('registered_contracts', sa.Column('blockchain_id', sa.UUID(), nullable=True))
+    op.create_foreign_key(op.f('fk_registered_contracts_blockchain_id_blockchains'), 'registered_contracts', 'blockchains', ['blockchain_id'], ['id'], ondelete='CASCADE')
+
+    # Manual - Start
+    op.execute(f"INSERT INTO blockchains (id, name, chain_id, testnet) VALUES ('{str(uuid.uuid4())}', 'polygon', 137, FALSE);")
```
zomglings, line 41: Can we do a single `INSERT INTO blockchains (...) VALUES` and add all blockchains in a single statement?

Can we add all blockchains we currently support? Ethereum Mainnet, Gnosis chain, etc.?

engineapi/engineapi/contracts_actions.py
```diff
@@ -66,30 +114,39 @@ def validate_method_and_params(
...
             raise InvalidAddressFormat("Parameter signer must be a valid address")
-        required_params["amount"] = str(required_params["amount"])
     else:
         raise ValueError(f"Unknown contract type {contract_type}")
 
 
 def register_contract(
     db_session: Session,
-    blockchain: str,
+    blockchain_name: str,
```
zomglings, line 123: Do we want to add lookup by chain ID as well? In that case, `blockchain_name` and `chain_id` would be `Optional` but not both can be `None`.

engineapi/engineapi/models.py
```diff
@@ -155,14 +157,87 @@ class DropperClaimant(Base):  # type: ignore
...
+
+
+class MetatxHolder(Base):  # type: ignore
+    """
+    MetatxHolder represents id of user from bugout authorization.
+    """
+
+    __tablename__ = "metatx_holders"
+
+    id = Column(
```
zomglings, line 191: Is this the `moonstream_user_id`? If so, we shouldn't put a `default`.

engineapi/engineapi/routes/metatx.py
```diff
@@ -37,7 +37,9 @@
 whitelist_paths = {
     "/metatx/openapi.json": "GET",
     f"/metatx/{DOCS_TARGET_PATH}": "GET",
+    "/metatx/blockchains": "GET",
     "/metatx/contracts/types": "GET",
```
zomglings, line 41: We should remove `metatx/contracts/types`?

engineapi/engineapi/models.py
```diff
@@ -155,14 +157,87 @@ class DropperClaimant(Base):  # type: ignore
...
+    description = Column(String, nullable=True)
+
+    call_requests = relationship(
+        "CallRequest",
+        back_populates="call_request_type",
+        cascade="all, delete, delete-orphan",
+    )
+
+
+class MetatxHolder(Base):  # type: ignore
```
zomglings, line 184: I think it's better to call this `MetatxRequester` not `MetatxHolder` - they are not holding anything, they are making a request to someone else to execute a function on a smart contract.

engineapi/engineapi/routes/metatx.py
```diff
@@ -295,7 +356,7 @@ async def create_requests(
     try:
         num_requests = contracts_actions.request_calls(
             db_session=db_session,
-            moonstream_user_id=request.state.user.id,
+            metatx_holder_id=request.state.user.id,
```
zomglings, line 359: Shouldn't there be a request type here?

zomglings (commented):
@kompotkot : some questions and suggestions

engineapi/alembic/versions/b4257b10daaf_call_request_types_and_metatx_holders.py
kompotkot: Added

engineapi/engineapi/contracts_actions.py
kompotkot: I suggest to add it in next versions, also we can modify get contract endpoints with chain id filter/testnet filter, etc

engineapi/engineapi/models.py
kompotkot: Done

engineapi/engineapi/routes/metatx.py
kompotkot: Saved for back capability, can delete it

engineapi/engineapi/models.py
kompotkot: Done

engineapi/engineapi/models.py
```diff
@@ -155,14 +157,88 @@ class DropperClaimant(Base):  # type: ignore
...
+
+    id = Column(
+        UUID(as_uuid=True),
+        primary_key=True,
+        default=uuid.uuid4,
+        unique=True,
+    )
+    request_type = Column(VARCHAR(128), nullable=False, unique=True, index=True)
+    description = Column(String, nullable=True)
+    required_params = Column(ARRAY(String), nullable=False)
```
zomglings, line 176: @kompotkot and I discussed this.

My feeling:
- Better to use JSON Schema for validation - https://json-schema.org/. Other request types (e.g. Ethereal mint) might have nested JSON structures inside the parameters object and `required_params` will quickly become obsolete as it only checks keys.
- Better to store schemas in code (in git). This way, all schema changes that make it to production can be reviewed and audited.
- We don't need a UUID primary key - we can just use the `request_type` as a primary key for this table. This will remove the need to read UUIDs from database in order to insert data into `call_requests`.

engineapi/engineapi/models.py
zomglings: Primary key: `user_id`?

engineapi/engineapi/contracts_actions.py
```diff
@@ -30,66 +37,136 @@ class InvalidAddressFormat(Exception):
...
+        expires_at=obj[0].expires_at,
+        created_at=obj[0].created_at,
+        updated_at=obj[0].updated_at,
+    )
+
+
 def validate_method_and_params(
-    contract_type: ContractType, method: str, parameters: Dict[str, Any]
-) -> None:
+    call_request_type: str,
```
zomglings, line 103: Let's make `call_request_type: Optional[str]`.

If `call_request_type is None`, then let us use `dropper-v0.2.0` as the default request type.

This will allow us to roll out these updates without breaking experience for our existing users - all of them are using `dropper-v0.2.0`.

engineapi/engineapi/models.py
zomglings: JSON schema validation - create issue doesn't have to be implemented in this PR. We can use existing validation for now.

zomglings (commented):
A few comments after further review.

zomglings (approved):
lgtm

Tested in local development environment and API interactions seem to be preserved.

PR #910, opened 2023-08-29
Add zksync era model.

Andrei-Dolgolev:
Commands:
Upgrade:
```
 ./alembic.sh -c configs/alembic.dev.ini upgrade head
```
Downgrade
```
./alembic.sh -c configs/alembic.dev.ini downgrade -1
```


kompotkot (approved):
lg

PR #916, opened 2023-08-31
Add initiate leaderboard config managment.

Andrei-Dolgolev:


engineapi/engineapi/actions.py
```diff
@@ -1491,6 +1505,130 @@ def list_leaderboards_resources(
...
+
+    if "status:active" not in entry.tags:
+        new_config["leaderboard_auto_update_active"] = False
+    else:
+        new_config["leaderboard_auto_update_active"] = True
+
+    return new_config
+
+
+def activate_leaderboard_config(
```
kellan-simiotics, line 1588: It seems a bit odd to check for only "status:active". Presumably someone could think it's correct to set "status:inactive". I suppose if we just tell everyone to use the API to update the config it will be fine.

engineapi/engineapi/settings.py
```diff
@@ -198,3 +198,11 @@
 MOONSTREAM_ADMIN_ID = os.environ.get("MOONSTREAM_ADMIN_ID", "")
 if MOONSTREAM_ADMIN_ID == "":
     raise ValueError("MOONSTREAM_ADMIN_ID environment variable must be set")
+
+MOONSTREAM_LEADERBOARD_GENERATOR_JOURNAL_ID = os.environ.get(
```
kellan-simiotics, line 202: Can we call this MOONSTREAM_LEADERBOARD_CONFIGURATION_JOURNAL_ID?

kellan-simiotics (changes requested):

engineapi/engineapi/actions.py
zomglings: Only the Moonstream API and our team has access to that journal. Moonstream users do not have access to the journal interface. Their only access is through this API.

engineapi/engineapi/settings.py
zomglings: +1

Andrei-Dolgolev: changed

kellan-simiotics (approved):

PR #922, opened 2023-09-11
Add Shadowcorns.

Andrei-Dolgolev:


crawlers/mooncrawl/mooncrawl/metadata_crawler/cli.py
```diff
@@ -55,7 +55,10 @@ def crawl_uri(metadata_uri: str) -> Any:
         try:
             response = urllib.request.urlopen(metadata_uri, timeout=10)
 
-            if response.status == 200:
+            if (
+                metadata_uri.startswith("data:application/json")
```
zomglings, line 59: `urllib.request.urllopen` correctly handles DATA URIs?

Does it also base64 decode the part after `data:application.json;`?

Andrei-Dolgolev: Yes but then response not have status itself

zomglings (approved):
lgtm

PR #924, opened 2023-09-20
Crawlers improvements.

Andrei-Dolgolev:
Added fix: 
eventJobs use deduplication by selector instead of event hash. 
Functions calls get selector from entry tags.


Andrei-Dolgolev:
Depends on https://github.com/moonstream-to/api/pull/919

PR #931, opened 2023-10-03
Field `live_at` for `call_requests`

kompotkot:
Resolves https://github.com/moonstream-to/api/issues/928
Resolves https://github.com/moonstream-to/api/issues/921


engineapi/alembic/versions/4f05d212ea49_live_at_for_metatx.py
```diff
@@ -0,0 +1,28 @@
...
+# revision identifiers, used by Alembic.
+revision = '4f05d212ea49'
+down_revision = '040f2dfde5a5'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column('call_requests', sa.Column('live_at', sa.DateTime(timezone=True), server_default=sa.text("TIMEZONE('utc', statement_timestamp())"), nullable=False))
```
zomglings, line 21: `live_at` should be nullable. Default can be `NULL`.

engineapi/engineapi/middleware.py
```diff
@@ -59,30 +143,33 @@ async def dispatch(
...
-                )
-                return Response(
-                    status_code=403,
-                    content="Only verified accounts can access journals",
-                )
-            if str(user.application_id) != str(MOONSTREAM_APPLICATION_ID):
-                return Response(
-                    status_code=403, content="User does not belong to this application"
-                )
+            user: BugoutUser = bugout_auth(token=user_token)
```
zomglings, line 162: I would expect to see an `if` statement like:

```python
if user_token is None:
    return call_next(request)
```

zomglings: Instead of letting the `bugout_auth` call fail.

engineapi/engineapi/models.py
```diff
@@ -315,6 +315,7 @@ class CallRequest(Base):
     parameters = Column(JSONB, nullable=False)
 
     expires_at = Column(DateTime(timezone=True), nullable=True, index=True)
+    live_at = Column(DateTime(timezone=True), server_default=utcnow(), nullable=False)
```
zomglings, line 318: I think it should be `nullable=true`. If `live_at = NULL`, then we always display that `call_request`. What do you think?

engineapi/engineapi/middleware.py
kompotkot: It could not be node, it could `""` and it will fall if it is empty at:
https://github.com/moonstream-to/api/blob/d3effd952cd0795883e5f5799dca2fff9074ac64/engineapi/engineapi/middleware.py#L62C21-L62C21

engineapi/engineapi/models.py
kompotkot: It will require to add additional logic which I would prefer to not do if the result will be same - just return list of call_requests

engineapi/alembic/versions/4f05d212ea49_live_at_for_metatx.py
kompotkot: It will require to add additional logic which I would prefer to not do if the result will be same - just return list of call_requests

engineapi/engineapi/middleware.py
zomglings: So put that condition also in the if statement?

kompotkot: It uses in `user: Optional[BugoutUser] = Depends(user_for_auth_header),` at line 302 in `metatx.py` where I handle user = `None` according with documentation [FastAPI Deps](https://fastapi.tiangolo.com/tutorial/dependencies/) it does not shows options where it `return call_next(request)` I think it is unsupported approach to use in FastAPI Dependencies. I can add
```python
if user_token != "":
  ...
else:
  return None
```
for clarification

engineapi/engineapi/contracts_actions.py
```diff
@@ -507,6 +517,21 @@ def list_call_requests(
             CallRequest.expires_at > func.now(),
         )
 
+    # If user id not specified, do not show call_requests before live_at.
+    # Otherwise check show_before_live_at argument from query parameter
+    if metatx_requester_id is not None:
+        query = query.filter(
+            CallRequest.metatx_requester_id == metatx_requester_id,
+        )
+        if not show_before_live_at:
```
Andrei-Dolgolev, line 526: Do it actualy must be out of that if?
Current logic kind of unintuitive if  set **show_before_live_at:** to true without `metatx_requester_id` it still will apply that filter on else.

PR #932, opened 2023-10-04
Complete call request

kompotkot:
Resolves https://github.com/moonstream-to/api/issues/930
Resolves https://github.com/moonstream-to/api/issues/929

For tests sig creation can be used tool: https://github.com/kompotkot/simple-signer

```
curl 'http://localhost:7191/metatx/requests/<request_id>/complete' \
--header 'Authorization: metatx <signer_base64_message>' \
--form 'tx_hash="<tx_hash>"'
```


kompotkot:
@bugout-dev check
- [ ] First merge https://github.com/moonstream-to/api/pull/931

engineapi/engineapi/routes/metatx.py
```diff
@@ -414,3 +431,32 @@ async def delete_requests(
         raise EngineHTTPException(status_code=500)
 
     return deleted_requests
+
+
+# @app.post("/requests/{request_id}/complete", tags=["requests"])
+@app.post("/requests/complete", tags=["requests"])
```
zomglings, line 437: This should be `/requests/<request_id>/complete` IMO

engineapi/engineapi/routes/metatx.py
```diff
@@ -414,3 +431,32 @@ async def delete_requests(
...
+async def complete_call_request_route(
+    tx_hash: str = Form(...),
```
zomglings, line 439: I think we should define a JSON body for the request.

And a JSON response type, as well.

Current setup is inconsistent with the rest of our API.

engineapi/engineapi/routes/metatx.py
```diff
@@ -414,3 +431,32 @@ async def delete_requests(
...
+    call_request_id: UUID = Form(...),
+    message=Depends(metatx_sign_header),
```
zomglings, line 441: Where is the message being used?

Shouldn't `complete_call_request` take the `message["caller"]` as an argument? Otherwise, how does this prevent some random person from signing a message using their own account and completing some other request?

zomglings (changes requested):
I don't see a `caller` check on the `complete_call_request` endpoint. That seems to be a serious issue.

engineapi/engineapi/routes/metatx.py
kompotkot: The problem in `whitelist_paths`, dynamic `<request_id>` and our current implementation of authentification middleware. I tried to avoid making this over complicated changes so this is the only way to make this route public.

engineapi/engineapi/routes/metatx.py
kompotkot: Ok

engineapi/engineapi/routes/metatx.py
kompotkot: Fixed with commit https://github.com/moonstream-to/api/pull/932/commits/e9c46f7a5beb58f3d9aae9e0247664e67c9da16d

zomglings: Somewhere we should normalize `caller` into a checksum address.

kompotkot: It happens in `metatx_sign_header`

engineapi/engineapi/routes/metatx.py
zomglings: How complex would it be to update our `whitelist_paths` implementation?

engineapi/engineapi/version.txt
```diff
@@ -1 +1 @@
-0.0.7
+0.0.8
```
zomglings, line 1: Bump version.

engineapi/engineapi/routes/metatx.py
zomglings: You didn't make this change.

engineapi/engineapi/middleware.py
```diff
@@ -26,13 +33,170 @@
...
+            logger.error(f"Error parsing auth header: {str(e)}")
+            raise EngineHTTPException(status_code=500, detail="Internal server error")
+
+        if token != "":
+            user = brood_auth(token=token)
+
+    return user
+
+
+async def metatx_sign_header(
```
zomglings, line 150: `metatx_sign_header` is a confusing name - it seems like it is *signing* something. But actually, it is *verifying* a signature.

Maybe `metatx_verify_header`?

zomglings (commented):
A few comments.

Andrei-Dolgolev (approved):
lgtm

PR #937, opened 2023-10-18
Nb metamask support and small fixes

kompotkot:
Support metamask, fixed path url lose when working with infura


zomglings (approved):
lgtm

PR #984, opened 2023-12-12
Add batching leaderboard generator.

Andrei-Dolgolev:
Add batching upload scores to leaderboard.
test
```
leaderboards-generator leaderboards-generate --query-api-access-token $MOONSTREAM_TOKEN --leaderboard_id <id_of_leaderboard>(where amount of scores more then 20k)
```
Add cli params `--leaderboard-push-batch-size` and `--leaderboard-push-timeout-seconds`
Via parameter --leaderboard_id now can run any leaderboard config.

Logic:
Create_version - > upload in batch ->  make version public -> delete (`new_version -1`)


crawlers/mooncrawl/mooncrawl/leaderboards_generator/utils.py
```diff
@@ -101,3 +106,120 @@ def get_results_for_moonstream_query(
...
+        )
+        return
+
+    leaderboard_version_id = leaderboard_api_response.json()["version"]
+
+    ## second step push data to leaderboard version
+
+    leaderboard_version_push_api_url = f"{MOONSTREAM_ENGINE_URL}/leaderboard/{leaderboard_id}/versions/{leaderboard_version_id}/scores?normalize_addresses={leaderboard_config['normalize_addresses']}&overwrite=false"
+
+    chunks = chunk_data(data, chunk_size=batch_size)
```
zomglings, line 180: You can also use a generator comprehension here instead of defining `def chunk_data...` function:

```python
chunks = (data[i:i+batch_size] for i in range(0, len(data), batch_size))
```

See this for more details on generator comprehensions: https://python-reference.readthedocs.io/en/latest/docs/comprehensions/gen_expression.html

zomglings (approved):
lgtm

PR #1038, opened 2024-03-28
Switch from access-id to web3-uri, zksync sepolia and avalanche

kompotkot:
- Removed --access-id, now it understands blockchain type or web3-uri
- Avalanche support
- zksync era sepolia basic crawlers

Tested.


Andrei-Dolgolev (approved):
lgtm

PR #1062, opened 2024-04-30
Verify if requests exists before push new list

kompotkot:
If query `verify=true` provided with POST /requests endpoint, it first verify call_requests not presented in database, if so, then return request_ids


kellan-simiotics (approved):
lg

PR #1068, opened 2024-05-20
Add uniqiue indexes for tx_call/event under seer label.

Andrei-Dolgolev:
```
./alembic.sh -c alembic.prod.ini upgrade head
```


kompotkot (approved):
lg

PR #1131, opened 2024-09-27
Update tags.

Andrei-Dolgolev:


moonstreamapi/moonstreamapi/actions.py
```diff
@@ -1184,7 +1184,7 @@ def create_resource_for_user(
 def chekc_user_resource_access(
```
kompotkot, line 1184: chekc_user_resource_access -> check_user_resource_access

moonstreamapi/moonstreamapi/routes/subscriptions.py
```diff
@@ -109,12 +109,14 @@ async def add_subscription_handler(
             user_token=token,
         )
 
-        if not results:
+        if results is None:
             raise MoonstreamHTTPException(
                 status_code=403,
```
kompotkot, line 114: Better to type 404, and error "Not found" even user do not have access to it

moonstreamapi/moonstreamapi/settings.py
```diff
@@ -309,6 +309,8 @@
     "user_id",
     "address",
     "blockchain",
+    "customer_id",
```
kompotkot, line 312: What about journal entities changes (at spire). We added here reserved tags, but not modified the source. Why just not to use `required_fields` https://spire.bugout.dev/journals/docs#tag/entities/operation/create_journal_entity__journal_id__entities_post
?

moonstreamapi/moonstreamapi/routes/subscriptions.py
```diff
@@ -624,7 +638,7 @@ async def update_subscriptions_handler(
         f"{key}:{value}"
         for tag in subscription_required_fields
         for key, value in tag.items()
-        if key not in MOONSTREAM_ENTITIES_RESERVED_TAGS
+        if key not in MOONSTREAM_ENTITIES_RESERVED_TAGS and key != "instance_name"
```
kompotkot, line 641: Why do we need here check instance name, since `instance_name` already in reserved tags?

moonstreamapi/moonstreamapi/settings.py
Andrei-Dolgolev: Old one remain same. New one will have new tags

kompotkot (approved):
lg

PR #1134, opened 2024-10-16
Set schema for version table

kompotkot:
Required by mdb-v3-controller to run upgrade for fresh instances automatically.


Andrei-Dolgolev (approved):
lgtm

PR #1165, opened 2025-01-26
Controll cache with env vars

kompotkot:
Gives more controll for client cache lifitime and health check interval with environment variables.

Tested locally.


Andrei-Dolgolev (approved):
lgtm

